PR #6, opened 2015-01-06
Reindex models on changes

longhotsummer:
Subscribe to notifications on changed models and reindex when necessary.

This cleans up a lot of unused code in search.py. In particular, it makes more assumptions about what models to index based solely on the list of transforms in backend/transforms.py rather than relying on lots of lists of types and index names.

When running rebuild_db.py, automatic reindexing is disabled. The full index must be recreated.

We should drop the entire PMG index in ES after this change, because some data type names have changed (eg. `calls_for_comment` is now `call_for_comment`).


petrus-jvrensburg:
Cool. @j-norwood-young can you check this and merge if you're happy?

PR #14, opened 2015-01-16
Handle API exceptions such that new relic ignores non-errors.

longhotsummer:
Change APIException to inherit from HTTPException and allow new
relic to ignore many errors based on status code, such as 404.

This fixes all the spurious errors we get for favicon.ico.


petrus-jvrensburg:
Okay, that's deployed. Thanks.

PR #15, opened 2015-01-16
Construct index URLs for backend for local host

longhotsummer:
Flask will construct a full URL using the server name used in the request, we don't need to hard code that.

### DIFF
--- a/backend/views.py
+++ b/backend/views.py
@@ -17,8 +17,6 @@ import math
 from flask_security import current_user
 from flask_security.decorators import load_user
 
-API_HOST = app.config["API_HOST"]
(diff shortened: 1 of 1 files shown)


backend/views.py
```diff
@@ -235,9 +233,11 @@ def landing():
     List available endpoints.
     """
 
+    base_url = url_for('landing', _external=True)
```
petrus-jvrensburg: Could we drop this call to url_for, and rather just use `request.base_url`?

longhotsummer: Yes, didn't know about that. Nice!

PR #17, opened 2015-01-22
Improve 404 and non-API 500-level error handling.

longhotsummer:
When the API gives us a 404, send the same to the user and show a nice page.

Decently handle non-API 500 responses.


longhotsummer:
This may not actually prevent New Relic from jumping up and down about 404s, let me take another look. Midnight ain't the best time to write code.

longhotsummer:
Ok, this should work.

PR #18, opened 2015-01-24
Show an 'edit page' button for all content types.

longhotsummer:


petrus-jvrensburg:
Thanks for this. 
I like how you added the `frontend_url_format` attribute to the Admin classes. It makes sense to have the logic there.

longhotsummer:
Thanks. It fits into flask-admin's declarative approach to configuration.

PR #19, opened 2015-01-25
Change from less to sass, improve header styles

longhotsummer:
Move from less to sass, which has a pure-python compiler so that we don't need extra things for compiling assets, it just works.

* pull in bootstrap-sass (all of it)
* change header navigation links to look like links
* scale down header font sizes, requested by PMG


petrus-jvrensburg:
I'm not deploying this and #18 right now, because I'm on patchy internet, but if you're in a hurry, please go ahead and do so.

longhotsummer:
I'll do a push.

PR #21, opened 2015-01-27
Use url field for file items from the API.

longhotsummer:
This means others who consume the API can also find the links
to the files, and simplifies our handling of file attachments.


petrus-jvrensburg:
That makes sense. But I don't see any changes to `models.py` or `backend/views.py`. Where is the `STATIC_HOST` injected into the url field?

longhotsummer:
That was done in a previous merge, this merge just finishes off the dangling bits I missed. See https://github.com/Code4SA/pmg-cms-2/blob/master/backend/models.py#L340

petrus-jvrensburg:
Great, thanks!

petrus-jvrensburg:
that's deployed

PR #29, opened 2015-02-02
Api

petrus-jvrensburg:


backend/models.py
```diff
@@ -359,6 +360,7 @@ def get_code(self):
     def to_dict(self, include_related=False):
         tmp = serializers.model_to_dict(self, include_related=include_related)
         tmp['code'] = self.get_code()
+        tmp['uri'] = url_for('resource_list', resource='bill', resource_id=self.id, _external=True)
```
longhotsummer: I think it makes more sense to call this field `url` instead of `uri`, since it's the location of the entity, not the identity.

longhotsummer:
There's a lot of cut and paste of duplicated code here, which sucks, but because we're inconsistent with URL endpoint names (question_reply vs committee-meeting) it makes it harder to be generic.

Longer term we should borrow search.py's class-name-to-dash-name code and work this all out programatically.

But this is good for now, I think.

petrus-jvrensburg:
Okay cool. That should be good to go then ;)

PR #30, opened 2015-02-04
Viewport, description and keyword metatags

j-norwood-young:


longhotsummer:
Nice. I'm going to hold off on merging this for now because with the new viewport setting, it looks really poor on mobile because the header doesn't collapse at all.

j-norwood-young:
I've removed the viewport tag, wanted to keep the other meta tags in

j-norwood-young:
I'm not getting a menu collapse with or without the tag

longhotsummer:
> I'm not getting a menu collapse with or without the tag

Yup, that's the problem. That needs to be fixed before we enable the mobile-friendly viewport.

longhotsummer:
This has been done outside of this pull request.

PR #38, opened 2015-02-09
Make new bill details page look like the old one

longhotsummer:


longhotsummer:
Ok, this should be good to go.

PR #40, opened 2015-02-10
Fix toggling of collapse elements.

longhotsummer:
Currently visible collapse elements must have the 'in' attribute. We could make this cleaner using bootstrap events, but it works for now.

Fixes https://trello.com/c/dxEdWG5A/328-committee-about-content-is-not-visible


j-norwood-young:
It's a bit weird that the text only shows up after the animation, but I'm approving anyway because I don't think it's critical.

PR #47, opened 2015-06-18
Make header responsive

xybrnet:
@longhotsummer

### DIFF
--- a/pmg/static/resources/css/style.scss
+++ b/pmg/static/resources/css/style.scss
@@ -286,8 +286,9 @@ article.page {
 
 .footer-1 {
   background-color: darken(#ddd, 40%);
-  @media (min-width:768px) {
     margin-top: 60px;
+  @media (max-width:768px) {
+    margin-top: 30px;
   }
 }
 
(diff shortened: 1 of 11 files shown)


pmg/static/resources/css/style.scss
```diff
@@ -213,6 +234,10 @@ article.page {
   img {
     margin-bottom: 10px;
   }
+
+  @media (max-width:768px) {
+    margin-top: 10px;
+  }
 }
 
 .correct-this-page-footer {
```
longhotsummer: We need an additional margin-bottom here of about 20px.

longhotsummer: Alternatively, let's just give .footer-1 some margin-top. The idea is to stop the content from being flush against it.

longhotsummer:
This is looking great.

A small tweak which would be nice (but isn't critical) is to make form submission buttons (such as login, password reset, etc.) full width on mobile. It makes them clearer tap targets.

I generally add a help class such as this:

```css
.full-xs {
  @media (max-width: 767px) {
    width: 100%;
  }
}
```

longhotsummer:
Btw, I'm happy to merge this and keep iterating. There are no show-stoppers here.

xybrnet:
@longhotsummer I've made the changes.

PR #57, opened 2015-08-18
MP attendance API for MPs

longhotsummer:
@xybrnet
 
This also ensures the attendance records are ordered most recent first.


xybrnet:
:+1:

PR #58, opened 2015-09-07
New questions listing

xybrnet:
@longhotsummer


pmg/templates/_committee_question_detail.html
```diff
@@ -24,11 +24,15 @@ <h4 class="media-heading">
...
         {% endif %}
+        to ask the
+        {{ question.question_to_name }}
       </h4>
 
-      <h5>{{ question.intro }}</h5>
-
-      <p>{{ question.question }}</p>
+      <div class="row">
+        <div class="col-xs-12 col-sm-9">
```
longhotsummer: You don't need col-xs-12 here. col-xs-12 just sets the width to 100%, which isn't necessary since a div tag is always display: block, width: 100%.

You basically never need col-*-12.

longhotsummer:
:+1: nice!

PR #61, opened 2015-09-09
Ministers

longhotsummer:
This begins support for Ministers and linking questions to ministers, not committees. It doesn't complete the whole picture, that's a larger change

* ministers exist as standalone things
* they're linked to committees that bear their name
* existing questions have been updated to be linked to a ministers AND a committee
* newly added questions will be automatically linked to the minister of the committee they're linked to

In time, we can remove the committee link and rely on the minister link.

@xybrnet


xybrnet:
:+1:

PR #62, opened 2015-09-10
FIX: allow meetings with attendance to be deleted.

longhotsummer:
Before this, when deleting a meeting with attendance the meeting_id of the attendance was set to null, which voliates the not-null constraint.

Now, when the meeting is deleted the attendance is too.

Sometimes SQLAlchemy confuses me.

@xybrnet


xybrnet:
:+1:

PR #63, opened 2015-10-06
Smaller search, clearer call to action.

longhotsummer:
@xybrnet

### DIFF
--- a/pmg/templates/_header.html
+++ b/pmg/templates/_header.html
@@ -2,7 +2,7 @@
   <div class="container">
     <div class="row">
 
-      <div class="col-xs-9 col-md-3 clearfix">
+      <div class="col-xs-9 col-md-4 clearfix">
         <a class="navbar-brand" href="/">
           <img id="logo" src="/static/resources/images/logo-small.png" alt="Parliamentary Monitoring Group" class="img-responsive">
           </a>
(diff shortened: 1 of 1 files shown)


pmg/templates/_header.html
```diff
@@ -55,12 +55,12 @@
...
 
-      <div class="col-xs-12 col-md-5 col-lg-6">
+      <div class="col-xs-12 col-md-4 col-lg-5">
         <form class="navbar-form" role="search" action="/search">
           <div class="input-group">
-            <input type="text" class="form-control" name="q" placeholder="Search for committee meetings, bills, ministers, schedules and documents" value="{{q}}">
+            <input type="text" class="form-control" name="q" placeholder="What are you looking for?" value="{{q}}">
             <span class="input-group-btn">
-              <input type="submit" class="btn btn-success" onclick="form.submit()" value="Search" />
+              <button type="submit" class="btn btn-success"><i class="fa fa-search"></i> Search</button>
```
xybrnet: What do you think about only displaying the search icon, with no text?

longhotsummer: I thought about that, and other sites do it, but I think being explicit is better. If Font Awesome doesn't load, then the user can't see what that button does at all.

xybrnet: Cool.

xybrnet:
:+1:

PR #65, opened 2015-10-06
Search emails

longhotsummer:
Internal support for doing a search and sending emails for updated content.

We still need to track modification dates for content.

@xybrnet


xybrnet:
:+1:

PR #66, opened 2015-10-06
Just use model.url for frontend urls in admin view.

longhotsummer:
Just a small simplification, we don't need the complex logic any more.

@xybrnet


xybrnet:
:+1:

PR #67, opened 2015-10-07
Bring DB up to date with models.

longhotsummer:
@xybrnet 

    This seems to be necessary because SQLAlchemy changed how it does
    foreign key constraints and uniqueness constraints. The net impact
    on the database is effectively zero.

    NOTE: this migration cannot go backwards because
    the names of some constraints are missing, see
    https://bitbucket.org/zzzeek/alembic/issues/331/autogenerated-opdrop_constraint-none-fails


xybrnet:
:+1:

PR #68, opened 2015-10-07
Track content changes

longhotsummer:
Track any change to resources exposed in the API and surface those in search results.

Currently ANY update to a resource will trigger this, including changing things other than the content. But I think that's okay for now.

@xybrnet


xybrnet:
Nice. I also think that will work fine for now.

:+1:

PR #69, opened 2015-10-07
Send email job

longhotsummer:
This queues up a task at 3am daily to send saved search emails. We'll need to test this in production once we have some alerts set up.

@xybrnet


xybrnet:
:+1:

PR #70, opened 2015-10-07
Send ajax request

xybrnet:
@longhotsummer


pmg/static/resources/javascript/pmg.js
```diff
@@ -12,6 +12,48 @@ $(function() {
...
+	        if (!/^(GET|HEAD|OPTIONS|TRACE)$/i.test(settings.type) && !this.crossDomain) {
+	            xhr.setRequestHeader("X-CSRFToken", csrftoken)
+	        }
+	    }
+	})
+
+	$(".create-alert").on("click", function(e) {
+
+		var q = $(this).data('q'),
+			  committee_id = parseInt($(this).data('committee')) || "",
```
longhotsummer: You don't need to worry about parsing ints, all form data gets sent as a string.

pmg/static/resources/javascript/pmg.js
```diff
@@ -12,6 +12,48 @@ $(function() {
...
+			$('.create-alert').addClass('hidden');
+			$('.remove-alert').removeClass('hidden').data('id', resp.id);
+		});
+	});
+
+	$(".remove-alert").on("click", function(e) {
+
+		var id = parseInt($(this).data('id')) || "";
+
+		$.post(
```
longhotsummer: Ugh, looks like making this a DELETE request is going to be a mission on both sides.

Let's do this instead, send a `POST` to `/user/saved-search/{id}/delete`

pmg/views.py
```diff
@@ -785,3 +793,26 @@ def correct_this_page():
         flash('Thanks for your feedback.', 'info')
 
     return redirect(request.form.get('url', '/'))
+
+
```
longhotsummer: I think we should have these in user_management.py

longhotsummer:
Nice :)

longhotsummer:
:+1:

PR #72, opened 2015-10-08
Apscheduler table

longhotsummer:
This ensures that alembic is happy, otherwise it keeps thinking the table is unnecessary.

@xybrnet


xybrnet:
:+1:

PR #73, opened 2015-10-08
Explicit sqlalchemy naming constraints

longhotsummer:
This fixes the issue in #67 that meant that that migration couldn't be run backwards.

@xybrnet


longhotsummer:
Wrong target branch.

PR #74, opened 2015-10-08
Specify naming constraints for SQLAlchemy so that alembic just works

longhotsummer:
Without this, Alembic doesn't generate names in a way that allows both
the upgrade and downgrade paths to work without manual changes.

This specifies naming constraints that match what SQLAlchemy has
used up until now.

This fixes the issue in #67 that meant that that migration couldn't be run backwards.

@xybrnet


xybrnet:
:+1:

PR #75, opened 2015-10-08
Improved stubs

longhotsummer:
There is still a horrible amount of duplication here, but this does bring most content types in line with each other.

@xybrnet


xybrnet:
:+1:

PR #81, opened 2015-10-13
Anonymous users search alerts

xybrnet:
Using a work around to set global vars in a template by appending to a list.

But I think the extra context within the remove-alert box doesn't work that well.

@longhotsummer

### DIFF
--- a/pmg/templates/_create_alert.html
+++ b/pmg/templates/_create_alert.html
@@ -1,18 +1,28 @@
+{% set filtered_committee = [] %}
+{% set filtered_data_type = [] %}
+
+{% for committee in committees %}
+  {% if filters.committee|int == committee.id|int %}
+    {% if filtered_committee.append(committee) %}{% endif %}
+  {% endif %}
+{% endfor %}
+
+{% for data_type, name in search_types %}
+  {% if filters.type == data_type %}
+    {% if filtered_data_type.append(name) %}{% endif %}
+  {% endif %}
+{% endfor %}
+
 <div class="row search-summary">
   <div class="col-sm-8">
     <h2 class="search-q">"{{ q }}"</h2>
+    {% if filtered_data_type %}
+      <p>in <b>{{ filtered_data_type[0] }}</b></p>
+    {% endif %}
 
-    {% for data_type, name in search_types %}
-      {% if filters.type == data_type %}
-        <p>in <b>{{ name }}</b></p>
-      {% endif %}
-    {% endfor %}
-
-    {% for committee in committees %}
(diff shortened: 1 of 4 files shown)


pmg/templates/_create_alert.html
```diff
@@ -23,15 +33,33 @@ <h2 class="search-q">"{{ q }}"</h2>
...
+          Get free email alerts when we add new information that match these search terms.
       </div>
 
       <div class="email-alert-box remove-alert {% if not saved_search %} hidden {% endif %}">
-        <a class="btn btn-warning" type="button" data-id="{{ saved_search.id }}">Remove alert</a>
-      </div>
+        <p>
+          <a class="btn btn-warning" type="button" data-id="{{ saved_search.id }}">Remove alert</a>
+        </p>
+          You are subscribed to <b>"{{ q }}"</b>
```
longhotsummer: This feels wordy, since the details of the search are just to the left of this box in any case.

Perhaps "We'll send you an email when we find new items for this search." or similar?

pmg/templates/_create_alert.html
```diff
@@ -23,15 +33,33 @@ <h2 class="search-q">"{{ q }}"</h2>
...
+      </div>
     </div>
   {% else %}
-    <p>Register for an account or log in to create alerts.</p>
+    <div class="col-sm-4">
+      <div class="email-alert-box create-alert-anon">
+        <p>
+          <a class="btn btn-default disabled" type="button" data-q="{{ q }}" data-type="{{ filters.type }}" data-committee="{{ filters.committee }}">Create alert</a>
+        </p>
+          <a href="{{ url_for('security.login', next=request.full_path) }}">Log in</a> or <a href="{{ url_for('security.register', next=request.full_path) }}"> create an account </a> to get free email alerts when we add new information that match these search terms.
```
longhotsummer: Shouldn't be whitespace around the content of the second <a> tag.

pmg/templates/_create_alert.html
```diff
@@ -1,21 +1,31 @@
+{% set filtered_committee = [] %}
```
longhotsummer: How about we create a helper object in the search method in pmg/views.py that is passed to the template and has metadata such as the committee name, committee id, content type, friendly content type, year etc. That'll really make the views simpler.

longhotsummer:
Nice, i like the links for the anonymous case.

I think creating the helper object for the template would simplify it greatly. I know it'd definitely make the mobile-friendly filter list easier.

longhotsummer:
Ok, based on PMG feedback, let's update the wording in the create search box.

PR #82, opened 2015-10-15
Manage email alerts

xybrnet:
@longhotsummer

### DIFF
--- a/pmg/user_management.py
+++ b/pmg/user_management.py
@@ -124,10 +124,9 @@ def email_alerts():
     else:
         subscriptions = set()
 
-    search_alerts = defaultdict(list)
-    saved_searches = current_user.saved_searches
-    for ss in saved_searches:
-        search_alerts[ss.search].append(ss)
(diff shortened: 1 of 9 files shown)


pmg/user_management.py
```diff
@@ -111,11 +112,12 @@ def email_alerts():
 
         # register a google analytics event
         ga_event('user', 'change-alerts')
-        flash("Your notification settings have been updated successfully.", "success")
-        if next_url:
-            return redirect(next_url)
 
-        return redirect(url_for('email_alerts'))
+        # When was this called?
+        # if next_url:
```
xybrnet: Do we need to continue supporting this? How would this then work?

longhotsummer: I think so, yes. When a user clicks the 'email alert' button on a committee page, it goes through this path. It would be nicer to do it via ajax without a page submit, but that's okay for now.

longhotsummer:
Ship it!

PR #83, opened 2015-10-19
FIX search alerts to improve results, exclude daily schedules

longhotsummer:
Filter recent results locally so that elasticsearch still returns
the top results. Otherwise, we only compare NEW results against
each other, which could have really poor matches.

@xybrnet


xybrnet:
:+1:
(Pending their response on excluding the daily schedule)

longhotsummer:
I've reverted the exclusion of Daily Schedules. We can revert the revert if necessary. I'm going to push this out.

PR #84, opened 2015-11-02
Retain filtered committee when searching from Q & R page

xybrnet:
@longhotsummer


longhotsummer:
That's it? Does it work? :) Pleasantly surprised.

xybrnet:
Unless I'm missing something. I was quite surprised too.

longhotsummer:
As long as it works.

PR #86, opened 2015-11-04
FIX alerts page for anonymous users

longhotsummer:
Show them the page so that they can see what they'll get,
but disable the controls and show a login/register box.

Also stop the form controls from firing on pages that aren't
the email alert page.

@xybrnet


longhotsummer:
These fixes are from errors reported by new relic.

xybrnet:
Nice :+1:

PR #87, opened 2015-11-04
Adjust elasticsearch to favour recent results.

### LINKED ISSUE #54: Sort search results by date
I am currently search for results in Hansards. I know that I am looking for a result in October 2009. I can filter by year but not by month. This means that I need to scan through the results. It would be much quicker if the results were sorted by date so that I can skip to the correct page in the results.

longhotsummer:
This applies a gaussian decay function to the
scores, starting when docs are 5 days old,
such that at 30 days old scores are 60% (0.6)
of what they'd normally be.

This will ensure we favour recent results without
biasing too heavily.

This will help to ensure that more recent, revelant
docs show up in search alerts, otherwise we
tend to have older docs in the top 20 results (simply
because there are more of them).

Fixes #54.

@xybrnet

### DIFF
--- a/pmg/search.py
+++ b/pmg/search.py
@@ -218,11 +218,11 @@ class Search:
                 "query": q,
                 "gauss": {
                     "date": {
-                        # Scores must decay, starting at docs from 5 days ago
+                        # Scores must decay, starting at docs from 7 days ago
                         # such that docs 30 days ago are at 0.6.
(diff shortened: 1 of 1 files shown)


longhotsummer:
More info here: https://www.elastic.co/blog/found-function-scoring

xybrnet:
:+1:

PR #88, opened 2015-11-11
Ensure member names and PA links are unique

longhotsummer:
This prevents duplicates which are confusing and break PA.org.za.

We've just gone through an exercise to clean this up. If your local data doesn't allow these migrations to run, just delete the offending MPs, or grab a latest copy of the production DB.


longhotsummer:
@xybrnet

xybrnet:
:+1:

PR #89, opened 2015-11-13
Support "quoted phrase" searches

longhotsummer:
We pick out quoted phrases and tell ES that they're required.
For non-phrase terms, we prefer if they're close together.

We update the highlighting to try to always highlight phrases.
It's a bit of a black art.

Finally, handle phrase searches that result in nothing, eg "   "

@xybrnet


xybrnet:
:+1:

PR #95, opened 2016-01-20
Use a recaptcha instead of a honeypot for 'correct this page' feedback

longhotsummer:
This is powered by Google's Recaptcha, I have added the creds to the environment.

@xybrnet


xybrnet:
:+1:

PR #97, opened 2016-01-21
Move actual start and end times to end of form

longhotsummer:
Gaile requested this a while ago. Flask Admin is weird, but
this seems to work ok.

@xybrnet


xybrnet:
That is a bit weird. :+1:

PR #99, opened 2016-01-22
Check if member is assigned to a party, return None if not

xybrnet:
The API endpoint was failing if no party is assigned to a member.
This checks if it is the case, and returns None is so.

@longhotsummer


longhotsummer:
:+1:

PR #102, opened 2016-02-24
API only returns members belonging to one of the three major parties

xybrnet:
@longhotsummer

### DIFF
--- a/pmg/api.py
+++ b/pmg/api.py
@@ -25,6 +25,10 @@ logger = logging.getLogger(__name__)
 
 api = Blueprint('api', __name__)
 
+# This is a temporary fix to only show attendance for members
+# of the three major parties until we determine how to present
+# faulty passed records for alternate members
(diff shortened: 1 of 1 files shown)


pmg/api.py
```diff
@@ -465,10 +465,18 @@ def committee_meeting_attendance_summary():
     """
     Summary of MP attendance of committee meetings.
     """
+
+    # This is a temporary fix to only show attendance for members
+    # of the three major parties until we determine how to present
+    # faulty passed records for alternate members
+    parties = ['ANC', 'DA', 'EFF']
```
longhotsummer: Perhaps put this either into a constant at the top of the file, or on the MemberAttendance class (or whatever it is). Otherwise I suspect we'll change the one down the line, but not the other.

xybrnet: Nice. I've placed it in a constant.

PR #103, opened 2016-02-25
Add Google utm codes to elements on homepage

xybrnet:
Not sure if this could be done better.
@longhotsummer

### DIFF
--- a/pmg/static/resources/javascript/pmg.js
+++ b/pmg/static/resources/javascript/pmg.js
@@ -81,7 +81,21 @@ $(function() {
       $target.addClass('expanded');
       $btn.addClass('show-less');
       $btn.text('Show less');
-      ga('send', 'event', 'show-more', 'click');
+      if ($target.hasClass('feature-cards')) {
+      	ga('send', 'event', 'featured', 'show-more');
+      } else {
+      	switch ($target.parent().attr('class').replace('subsection', '').trim()) {
+      		case 'committee-meetings':
+      			ga('send', 'event', 'committee-meetings', 'show-more');
+      			break;
+      		case 'bills':
+      			ga('send', 'event', 'bills', 'show-more');
+      			break;
+      		case 'questions':
+      			ga('send', 'event', 'questions', 'show-more');
+      			break;
+      	}
+      }
     }
   });
 });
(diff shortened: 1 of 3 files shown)


pmg/static/resources/javascript/pmg.js
```diff
@@ -81,6 +81,14 @@ $(function() {
       $target.addClass('expanded');
       $btn.addClass('show-less');
       $btn.text('Show less');
+      ga('send', 'event', 'show-more', 'click');
```
longhotsummer: I wonder if it would make sense to have ``(...'recent-committee-meetings', 'show-more')`` and ``(... 'featured', 'show-more')``.

pmg/templates/index.html
```diff
@@ -71,9 +72,10 @@ <h6 class="date">{{ meeting.date|pretty_date|nbsp|safe }}</h6>
         <div class="col-sm-6 col-sm-offset-3 text-center">
           <h2>Search the PMG website</h2>
 
-          <form role="search" action="/search">
+          <form class="search-form" role="search" action="/search" name="middle">
             <div class="input-group">
               <input type="text" class="form-control" name="q" placeholder="Search the PMG website">
+              <input name="ga" type="hidden" value="search_middle">
```
longhotsummer: You can use the same utm tags in forms, since this is a GET form, instead of having to use GA events.

```html
<input name="utm_source" type="hidden" value="homepage">
<input name="utm_medium" type="hidden" value="s-middle">
```

pmg/templates/index.html
```diff
@@ -46,7 +47,7 @@
           <div class="row">
             {% for meeting in featured_content.committee_meetings %}
               <div class="col-xs-12 col-sm-6 col-md-3">
-                <a href="{{ url_for('committee_meeting', event_id=meeting.id) }}">
+                <a href="{{ url_for('committee_meeting', event_id=meeting.id) }}?utm_source=homepage&utm_medium=feature">
```
longhotsummer: Nice. I think you can include extra params as part of ``url_for``, like this:

``url_for('committee_meeting', event_id=meeting.id, utm_source='homepage', utm_medium='featured')``

pmg/templates/index.html
xybrnet: I'm not sure I understand how this would work?

xybrnet: Ah, I think I've got it.

longhotsummer:
Cool!

PR #106, opened 2016-03-29
Sendgrid email

xybrnet:
This moves functionality around sending email away from Mandrill, to Sendgrid.

@longhotsummer


config/test/config.py
```diff
@@ -35,19 +35,16 @@
...
-MANDRILL_TRANSACTIONAL_SUBACCOUNT = 'transactional'
-MANDRILL_ALERTS_TEMPLATE = 'notification-template'
-MANDRILL_ALERTS_SUBACCOUNT = 'alerts'
+# Sendgrid
+SENDGRID_API_KEY = env.get('SENDGRID_API_KEY')
+SENDGRID_TRANSACTIONAL_TEMPLATE_ID = '18988dde-425f-4abc-a62f-36f9221e3ff1'
 
 # Flask-Mail
-MAIL_SERVER = 'smtp.mandrillapp.com'
+MAIL_SERVER = 'smtp.sendgrid.com'
```
longhotsummer: For the test config, I'd ensure this is a fake value to avoid sending emails when testing by mistake. Ditto for the username and password.

pmg/__init__.py
```diff
@@ -41,14 +43,31 @@
...
+            "templates": {
+                "settings": {
+                    "enable": "1",
+                    "template_id": app.config['SENDGRID_TRANSACTIONAL_TEMPLATE_ID']
+                }
+            },
+            "ganalytics": {
+                "settings": {
+                    "enable": "1",
+                    "utm_source": "some_source",
```
longhotsummer: double utm_source entry?

pmg/models/emails.py
```diff
@@ -151,46 +152,40 @@ def find_or_create(cls, user, q, content_type=None, committee_id=None):
...
+    message.add_filter('ganalytics', 'utm_campaign', utm_campaign)
+    message.add_filter('clicktrack', 'enable', '1')
+
+    message.set_html(html)
+    message.set_from(from_email)
+    status, msg = sg.send(message)
 
     log.info("Email will be sent to %d recipients." % len(recipients))
-    log.info("Sending email via mandrill: %s" % msg)
+    log.info("Sending email via sendgrid: %s" % msg)
```
longhotsummer: This should come before `sg.send`. Does %s on the `msg` object print anything meaningful?

pmg/__init__.py
xybrnet: Shoot. Fixed.

config/test/config.py
xybrnet: Cool, done.

pmg/models/emails.py
xybrnet: I've updated it so that it prints: '200 Success' once successfully sent.
But it needs to be after the send in order to have msg available.

longhotsummer: Sorry, I misread this. I'd like us to log that we're about to send the email just before we send it ("Email will be sent to..."). That provides context if the sg.send call raises an exception.

Then after it's sent we can log the result of the send.

Pls also remove the `print`.

PR #116, opened 2016-06-28
GA dimensions for content date and cte

longhotsummer:
We need to setup these dimensions in GA before doing this.


longhotsummer:
GA dimensions have been configured.

PR #131, opened 2016-08-31
Committee meeting page layout

longhotsummer:
Based on Peter's mockup

* simplify header (date, chairperson)
* tweak font sizes so that h1 isn't so big
* separate tabs for summary and full minutes to make navigation simpler


longhotsummer:
@smmbll this is now up to date with master. A breaking change is that it moves the committees.scss into styles.scss so that all the variables work (python's sass compiler is a bit lame).

I've merged in your changes from master, this can be merged if you're happy.

PR #132, opened 2016-08-31
Upload audio to soundcloud and present using embedded player when ready

jbothma:
- add app.py command to upload batches of tracks to soundcloud, then check what's ready at SC
- show audio links as before
- if a track is uploaded to soundcloud and ready, play it in embedded player when clicked
- if a track isn't ready on soundcloud, open in browser as before.
- after embedding one track, clicking a different track replaces the embed of the first clicked track
- remove special case for meeting 10724 because the track's missing but keep meeting 13370 for comparison and communicating presentation options with PMG/Paul for now. 
- doesn't handle failed tracks any more than setting their state to "failed" and presenting them as we do now. proper handling of failed tracks depends on how many we see when we do a bunch bunch of tracks to get a notion of failure rate and SC's response to https://www.soundcloudcommunity.com/soundcloud/topics/http-500-when-deleting-track-succeeds-anyway . right now it's not possible to re-upload tracks to the same track ID via the API, according to SC docs somewhere.

before clicking a soundcloud track link:
pmg-audio-unclicked

after clicking one or more soundcloud track link:
pmg-audio-clicked


pmg/sync_soundcloud.py
```diff
@@ -0,0 +1,61 @@
...
+from pmg.models import *
+from pmg.models.soundcloud import SoundcloudTrack
+from sqlalchemy import desc
+import soundcloud
+
+logger = logging.getLogger(__name__)
+
+
+def get_unstarted_query():
+    return db.session.query(File, Event.title.label('event_title')) \
```
longhotsummer: Just sanity check that the SQL under the hood is joining on the correct columns.

pmg/sync_soundcloud.py
```diff
@@ -0,0 +1,61 @@
...
+def sync():
+    q = get_unstarted_query()
+    logging.info("Audio files yet to be uploaded to SoundCloud: %d" % get_unstarted_count(q))
+    batch = get_unstarted_batch(q)
+    logging.info("Uploading %d files to SoundCloud" % get_unstarted_count(q))
+    for file, event_title in batch:
+        upload_file(file, event_title)
+
+
+def upload_file(file, event_title):
```
longhotsummer: I'd put most of the logic in this file onto the SoundcloudTrack class. It's directly related, @classmethods are useful ways of scoping logic into a module. eg. ``SoundcloudTrack.new_from_event_file(event_file)`` or something.

pmg/sync_soundcloud.py
```diff
@@ -0,0 +1,61 @@
...
+    logging.info("Downloading from S3: %s" % file)
+    file_handle = file.download()
```
longhotsummer: Is the downloaded file disposed of at some point?

pmg/models/soundcloud.py
```diff
@@ -0,0 +1,32 @@
...
+    modified_at = db.Column(
+        db.DateTime(timezone=True),
+        nullable=False,
+        default=datetime.utcnow
+    )
+    file_id = db.Column(
+        db.Integer,
+        db.ForeignKey('file.id', ondelete='CASCADE'),
+        nullable=False
+    )
```
longhotsummer: If you define a relationship then you can include a backref so that you could go `file.soundcloud_track`:

```
file = db.relationship('File', lazy='joined', backref='soundcloud_track')
```

pmg/models/soundcloud.py
```diff
@@ -0,0 +1,32 @@
...
+    uri = db.Column(db.String())
+    state = db.Column(db.String())
+
+    def __init__(self, file, track):
```
longhotsummer: You might want to use a factory method instead. SQLAlchemy defines a constructor for you which takes kwarg params matching the field/attribute names. I'm not sure how this would interact with it.

pmg/sync_soundcloud.py
jbothma: :+1:

pmg/sync_soundcloud.py
jbothma: yup - when closed, because of tempfile.NamedTemporaryFile(delete=True)

pmg/models/soundcloud.py
jbothma: :+1:

pmg/sync_soundcloud.py
jbothma: :+1:

longhotsummer:
Can we nuke the special cases completely and take the tracks that exist for those and add them to the DB?

PR #135, opened 2016-09-06
Added correct margins to ad-hoc committee h3

smmbll:


longhotsummer:
Not sure how this fits in, but if it's needed, then great.

PR #136, opened 2016-09-08
Changed committee view to two column layout

smmbll:


smmbll:
Changes here are present in committees_page_filter branch

PR #141, opened 2016-09-08
Committee page filter

smmbll:


pmg/views.py
```diff
@@ -244,7 +247,40 @@ def committees():
...
+    reg_committees = deepcopy(adhoc_committees)
+    committees_type = None
+
+    for committee in committees:
+        if committee['ad_hoc'] is True:
+            committees_type = adhoc_committees
+        else:
+            committees_type = reg_committees
+
+        if committee['house_id'] is 3:
```
jbothma: I'd suggest breaking these magic numbers out to constants like Committee.NAT_ASSEMBLY

pmg/static/resources/javascript/committees.js
```diff
@@ -1,4 +1,97 @@
...
+
+$committeeNavItem.on('click', function (e) {
   e.preventDefault();
   $(this).tab('show');
+  $committees = $('.committees-list .tab-pane.active .committee');
+
+  if(showingSearchResult) clearSearchResult();
 });
+
+$clearResults.on('click', function() {
```
longhotsummer: we can remove the 'clear results' elements and code, right?

pmg/static/resources/javascript/committees.js
```diff
@@ -1,4 +1,97 @@
...
+  clearSearchResult();
+  $clearResults.hide();
+});
+
+$searchInput.on('focus', function() {
```
longhotsummer: Why attach the keyup event on focus? Won't the keyup event handler be added multiple times?

pmg/static/resources/javascript/committees.js
```diff
@@ -1,4 +1,97 @@
...
+      } else {
+        $searchResult.find('.left ul')
+          .append($results);
+      }
+    } else {
+      $resultsFound.hide();
+      $noResults.show();
+    }
+  }
+},500);
```
longhotsummer: I think this should be a bit lower, say 200. There's a brief pause where it looks like nothing has happened. It's not like what it's doing is particularly intensive.

longhotsummer:
Nice, this is looking great! Just some minor comments.

PR #144, opened 2016-09-13
Reworked bottom nav of committee detail page

longhotsummer:


pmg/static/resources/css/style.scss
```diff
@@ -946,6 +958,34 @@ body.email_alerts {
   max-height: 150px;
 }
 
+.c-m-nav {
```
longhotsummer: I think this is a bit too short to be meaningful. Perhaps cte-meeting-nav if committee is too long?

pmg/static/resources/css/style.scss
```diff
@@ -1050,3 +1090,23 @@ body.email_alerts {
     }
   }
 }
+
+@media screen and (max-width:544px) {
```
longhotsummer: I suggest using bootstrap's .visible-xs-block (or just .visible-xs if we're a few versions behind) - http://getbootstrap.com/css/#responsive-utilities

That saves us from writing the bulk of this css.

Also .btn-block makes buttons 100% wide.

pmg/static/resources/javascript/committees.js
```diff
@@ -86,3 +86,61 @@ $committeeNavItem.on('click', function (e) {
...
+      $rows.each(function(i,row) {
+        var $row = $(row);
+        var date = new Date($row.attr('data-date'));
+
+        if(currentDate.getMonth() - date.getMonth() > 6) $row.hide();
+      });
+      break;
+  }
+
+  $table.fadeIn({ duration: 250, easing: 'linear' });
```
longhotsummer: Nice touch!

pmg/templates/committee_detail.html
```diff
@@ -16,73 +16,98 @@ <h3 class="collapse-link"><span class="fa fa-caret-right"></span> <a href="#">Ab
...
+          <li><a href="#cc" data-toggle="tab">Calls for comment</a></li>
+          <li><a href="#tr" data-toggle="tab">Tabled reports</a></li>
+          <li><a href="#qr" data-toggle="tab">Questions/replies</a></li>
+        </ul>
+      </nav>
+
+      <div class="c-m-mobile">
+        <nav class="c-m-nav-mobile" id="c-m-nav-mobile">
+          <div class="dropdown">
+            <button type="button" class="btn btn-default" data-toggle="dropdown">Sections <span class="carat"></span></button>
```
longhotsummer: How horrible does it look if we just use a <select> element here? Right now, the dropdown doesn't change to indicate what is the currently chosen selection. If we use a <select> and it looks ok, it'd do that automatically.

pmg/views.py
```diff
@@ -210,6 +210,14 @@ def committee_detail(committee_id):
     """
     logger.debug("committee detail page called")
     committee = load_from_api('committee', committee_id)
+    YEAR_CUTOFF = 2012
```
longhotsummer: Let's sanity check this. We could also ask elasticsearch to give us stats... hmmm.

pmg/views.py
```diff
@@ -218,10 +226,24 @@ def committee_detail(committee_id):
     recent_questions = load_from_api(
         'minister-questions-combined',
         params=params)['results']
+    all_meetings = filter(lambda m: m['type'] == 'committee-meeting' and format_date(m['date']).year >= YEAR_CUTOFF,committee['events'])
```
longhotsummer: A more pythonic way of doing this is to use a list comprehension, which is generally preferred to using filter:

```python
all_meetings = [m for m in committee['events'] if m['type'] == 'committee-meeting' and format_date(m['date']).year >= YEAR_CUTOFF]
```

longhotsummer:
This is looking great, it's a much nicer experience.

The members and contact details column should still be on the left, I think.

pmg/views.py
```diff
@@ -210,6 +210,14 @@ def committee_detail(committee_id):
     """
     logger.debug("committee detail page called")
     committee = load_from_api('committee', committee_id)
+    YEAR_CUTOFF = 2012
+    now = datetime.now()
+    meetings_by_year = {}
+
+    def format_date(m_date):
```
longhotsummer: This seems weird to me. What if we just grabbed the first 4 chars from the date string and used that as the year?

pmg/templates/committee_detail.html
```diff
@@ -92,16 +117,14 @@ <h3 class="collapse-link"><span class="fa fa-caret-right"></span> <a href="#">Ab
...
-          </div>
-        {% endif %}
-
-        {% if recent_questions %}
-          <div class="col-md-6">
-            <h3><span class="fa fa-question-circle"></span> Questions and Replies</h3>
-            <table class="table table-compact">
+          {% endif %}
+        </div>
+        <div class="tab-pane fade in c-m-q-replies" id="qr">
```
longhotsummer: Do these tabs need both the id and the c-m-... class?

pmg/static/resources/javascript/committees.js
```diff
@@ -86,3 +86,61 @@ $committeeNavItem.on('click', function (e) {
...
+    case 'recent':
+      $rows.slice(10)
+        .hide();
+      break;
+    case 'six-months':
+      $rows.each(function(i,row) {
+        var $row = $(row);
+        var date = new Date($row.attr('data-date'));
+
+        if(currentDate.getMonth() - date.getMonth() > 6) $row.hide();
```
longhotsummer: This will break across different years, won't it?

longhotsummer:
I'm trying to think if we can simplify how we filter based on the tab selection. Currently we do some in python (grouping by year) and some in javascript (recent). Could we do all of it in one place? Either server side or on the client? That would save you duplicating code/logic.

pmg/views.py
smmbll: Yeah, I chose this arbitrarily--it was in line with the wireframe, but there's no particular reason why we'd stop there. We have data going back a very long time; I think I remember committee meetings dated to the mid-90's.

pmg/templates/committee_detail.html
smmbll: They need the id in order to be targeted by bootstrap tabs. The class is there in case any additional styling changes need to be made.

pmg/views.py
```diff
@@ -226,24 +226,25 @@ def get_year_unicode(m_date):
...
 
-        if meeting_year not in meetings_by_year:
-            meetings_by_year[meeting_year] = []
+        if meeting_year not in filtered_meetings:
+            filtered_meetings[meeting_year] = []
+
+        filtered_meetings[meeting_year].append(meeting)
 
-        meetings_by_year[meeting_year].append(meeting)
+    filtered_meetings['six-months'] = [m for m in all_meetings if (now.month - get_month_unicode(m['date']) <= 6) and (get_year_unicode(m['date']) == now.year)]
```
longhotsummer: This ain't gonna work for the first few months of the year (it'll ignore everything in the previous year).

Perhaps [parse the event date](http://stackoverflow.com/questions/969285/how-do-i-translate-a-iso-8601-datetime-string-into-a-python-datetime-object) and the just use ``(now - event_data).days < 30*6``?

app.py
```diff
@@ -3,6 +3,8 @@
 from flask.ext.script import Server, Manager
 from flask.ext.migrate import MigrateCommand
 
+app.debug = True
```
longhotsummer: You shouldn't need this, it's set to True in the config file that gets loaded, unless I'm missing something.

pmg/static/resources/javascript/committees.js
```diff
@@ -86,3 +96,55 @@ $committeeNavItem.on('click', function (e) {
...
+  var $selected = $('option:selected',this);
+  $selected.tab('show');
+  $cdNavItem.closest('li')
+    .removeClass('active')
+    .has('[data-target="' + $selected.attr('data-target') + '"]')
+    .addClass('active');
+});
+
+// Hide all but first ten meetings on load
+$cdCurrentYearTable.find('tr')
```
longhotsummer: Since the first tab to load is the 'last 3 months' or whatever, I don't think we need to do this?

pmg/static/resources/css/style.scss
```diff
@@ -936,6 +936,18 @@ body.email_alerts {
 }
 
 // committee detail page
+.committee_detail {
+  .tab-content {
+    padding:15px;
```
longhotsummer: can we make this ``padding-top: 15px``? Otherwise on mobile we're losing useful spaces on the sides. On desktop I'm ambivalent.

pmg/templates/committee_detail.html
```diff
@@ -38,35 +32,41 @@ <h3 class="collapse-link"><span class="fa fa-caret-right"></span> <a href="#">Ab
...
-                {% if recent_questions %}
                 <option data-target="#qr" data-toggle="tab" value="#qr">Questions/replies</option>
-                {% endif %}
               </select>
             </div>
           </nav>
           <div class="tab-pane fade in active" id="cm">
             <div class="cte-meetings-filter-outer hidden-xs">
               <div class="btn-group cte-meetings-filter-inner">
+                {% if filtered_meetings['six-months'] | length > 0 %}
```
longhotsummer: The first one to be shown needs to have the 'active' class on the button and the tab-pane.

pmg/static/resources/javascript/committees.js
smmbll: We need it for the mobile/desktop transition. If you select something on one, it needs to be selected on the other automatically.

PR #152, opened 2016-09-19
Added committee meetings search, removed committee meeting tabs

smmbll:


longhotsummer:
Nice! This is getting better and better.

A few items:

* the `<i>` tag around the Premium on the committee detail page isn't closed correctly, so the Premium is rendered in the wrong font
* when the cte detail page loads for the first time, the year dropdown box doesn't have anything selected, so no meetings show up
* on the calls for comment tab: let's directly write the HTML to show those, since currently we're using col's inside a TD, which is just weird and messes up the spacing
* the moment I typed more than one word on the filter, I realised it was doing literal filtering rather than keyword. Either we can do keyword, or we can just use [lunr](http://lunrjs.com/) to do the client-side filtering (which is actually search). That'll do stemming and keywords, and is dead-simple.
* tabled reports should now all have a `start_date`, let's put that into a column so users can see the date

The "about this committee" link is now totally getting in the way, grr. But we can fix that in a little bit.

pmg/static/resources/javascript/committees.js
```diff
@@ -54,70 +59,50 @@ var debounce = function(func, wait, immediate) {
...
 
-    resultsCount = $results.length;
+  if(!!resultsCount) {
+    $resultsList.show();
 
-    if(!!resultsCount) {
-      $resultsList.show();
+    if(!!params && params.sort) {
+      results.sort(function(a,b) {
+        return new Date($(b).find('.date').text()) - new Date($(a).find('.date').text());
```
longhotsummer: If the date strings are 'yyyy-mm-dd', then just sorting the text strings is the same as converting to dates and sorting. So you don't strictly need to convert to dates.

longhotsummer:
Great. The items above still need to be addressed, please. lunr.min.js also needs to be added to pmg/__init__.py around line 105, otherwise it isn't served to the user.

longhotsummer:
Great, ship it!

PR #160, opened 2016-10-17
Put party affiliations back into committee detail page

smmbll:


longhotsummer:
Make Parties Great Again!

PR #163, opened 2016-10-20
FIX table font size bug, which was due to bad html

longhotsummer:
@guushoekman @smmbll


longhotsummer:
Can I get a hell yeah on this?

guushoekman:
hell yeah!

PR #164, opened 2016-10-20
Embedded soundcloud audio directly into meetings page

smmbll:


pmg/templates/macros/attachments.html
```diff
@@ -29,16 +29,18 @@ <h4 class="panel-title">Audio</h4>
...
           {% for item in items %}
           <li>
-            <i class="fa fa-li fa-headphones"></i>
-            <a href="{{ item.url }}"
-               target="_blank"
-               data-soundcloud-uri="{{ item.soundcloud_uri }}"
-               class="audio">
-              {{ item.title or item.description or item.origname or item.file_path }}
-            </a>
+            <audio controls>
```
longhotsummer: This isn't the experience we want to go with, primarily because the recordings are very long and this doesn't allow the user to jump into it mid-way easily (they have to download everything before, first).

We just need to make the soundcloud embed show by default for the first audio recording that has soundcloud info linked to it.

longhotsummer: Guus and I just tested this on Safari, FF and Chrome. Chrome is by far the best, but Safari doesn't allow the user to jump into the audio. FF just doesn't load at all. :(

PR #166, opened 2016-11-01
User following

guushoekman:


pmg/templates/_top_links.html
```diff
@@ -1,5 +1,46 @@
-<li><a href="/committees/">Committees</a></li>
-<li><a href="/members">MPs</a></li>
-<li><a href="/bills">Bills</a></li>
-<li><a href="/question_replies">Questions and Replies</a></li>
-<li><a href="/calls-for-comments">Calls for Comments</a></li>
+<div class="menu-container hidden-xs">
```
jbothma: this should be inside an li - onlu li's can be children of ul's https://stackoverflow.com/questions/11755628/can-i-use-div-as-a-direct-child-of-ul

pmg/static/resources/css/style.scss
```diff
@@ -42,6 +42,10 @@ $yellow: #F5A623;
   font-family: $headings-font-family;
 }
 
+table {
```
longhotsummer: This is not necessary, we've fixed what caused this. Please make sure you're up to date from master.

pmg/templates/_top_links.html
```diff
@@ -1,5 +1,46 @@
...
+    <div class="menu">
```
longhotsummer: Aaargh! My eyes! :) Use 2 spaces, not tabs pls.

pmg/templates/layout.html
```diff
@@ -44,6 +44,7 @@
   {% assets "js" %}
     <script type="text/javascript" src="{{ ASSET_URL }}"></script>
   {% endassets %}
+  <script type="text/javascript" src="/static/resources/javascript/megamenu.js"></script>
```
longhotsummer: Please add this to assets.py, not here. That ensures it's included in the versioned asset bundle, is cached by the client, etc.

pmg/views.py
```diff
@@ -76,6 +76,27 @@ def classify_attachments(files):
...
+    if current_user.is_authenticated():
+
+        # Append user-followed committees if logged in
+        user_following = current_user.following
+
+        for committee in user_following:
+            meetings = load_from_api('v2/committees/%s/meetings' % committee.id,
+                fields=['id','title','date'])['results']
+
+            if len(meetings):
```
longhotsummer: `if meetings` is more pythonic.

pmg/views.py
```diff
@@ -76,6 +76,27 @@ def classify_attachments(files):
...
+                setattr(committee, 'recent_meeting', meetings[0])
```
longhotsummer: Just use `committee.recent_meetings = meetings[0]`

pmg/views.py
```diff
@@ -252,8 +274,8 @@ def get_month_unicode(m_date):
 
         filtered_meetings[meeting_year].append(meeting)
 
-    latest_year = max(y for y in filtered_meetings)
-    earliest_year = min(y for y in filtered_meetings)
+    latest_year = max(y for y in filtered_meetings) if len(filtered_meetings) else None
```
longhotsummer: In python, `if filtered_meetings` is generally how this test is done. Both None and an empty list are falsey.

longhotsummer:
Don't forget to merge in master to resolve the merge conflict. :)

pmg/views.py
smmbll: can't do that -- committee is a sqlalchemy object, which cannot be assigned to directly

pmg/templates/layout.html
guushoekman: Maybe I'm blind but where is assets.py?

longhotsummer: Oops, wrong project. Check pmg/__init__.py line 104.

longhotsummer:
It looks like this commit is missing a migration file to create the new tables?

pmg/views.py
longhotsummer: Try it out -- in python you can assign properties to any object.

pmg/user_management.py
```diff
@@ -64,6 +64,33 @@ def user_committee_alert(committee_id):
 
     return redirect(request.values.get('next', '/'))
 
+@app.route('/user/follow/committee/<int:committee_id>', methods=['POST'])
+def user_follow_committee(committee_id):
+    if current_user.is_authenticated() and request.method == 'POST':
+        u = current_user.follow_committee(Committee.query.get(committee_id))
```
longhotsummer: This code no longer works, because `follow_committee` doesn't return anything. SQLAlchemy knows how to track changed objects, you can simply do this:

```python
current_user.follow_committee(Committee.query.get(committee_id))
db.session.commit()
```

pmg/user_management.py
```diff
@@ -64,6 +64,33 @@ def user_committee_alert(committee_id):
...
+
+        if u is not None:
+            db.session.add(u)
+            db.session.commit()
+            ga_event('user','follow-committee','cte-follow-committee')
+        else:
+            flash('Committee %s cannot be followed.' % committee_id)
+
+    return redirect(request.values.get('next', '/'))
```
longhotsummer: I suggest this instead, which ensures that the user goes back to their request paging if AJAX fails for some reason. We're not using the `next` param in any case.

```
return redirect(request.headers.get('referer', '/'))
```

pmg/models/users.py
```diff
@@ -146,6 +149,18 @@ def gets_alerts_for(self, committee):
...
+        from ..models.resources import Committee
+        if not isinstance(committee, Committee):
+            committee = Committee.query.get(committee)
+        self.following.append(committee)
+
+    def unfollow_committee(self, committee):
+        from ..models.resources import Committee
+        if not isinstance(committee, Committee):
+            committee = Committee.query.get(committee)
+        self.following.remove(committee)
```
longhotsummer: The idea is that if you aren't following a cte, then you won't get alerts either. So here we should also remove the cte from the `committee_alerts` list.

longhotsummer:
In terms of UI, we need to ensure a good experience if the user isn't following any committees. We could show instead the top 10 most followed committees, and an call to action to change them.

I think we also need to tweak the colours in the megamenu. There's not much room and too much colour looks quite busy. Could we make the bulk of it (all of it?) white and rely on headings and consistent whitespace?

longhotsummer:
Please check the /email-alerts/ page, it's borked. The styling for `input`, `form` and `label` seems far too broad to me, it's going to have unintended side effects. I suggest putting a class on the forms that require that styling and hang the styles off of that.

pmg/views.py
smmbll: returns an error

smmbll: http://stackoverflow.com/questions/10588375/can-i-assign-values-in-rowproxy-using-the-sqlalchemy

as you can see there, a typical pattern is to convert the object to a dict first.

src/pip-delete-this-directory.txt
```diff
@@ -0,0 +1,5 @@
+This file is placed here by pip to indicate the source was put
```
longhotsummer: Pls be careful with `git add -A` or `git commit -A` -- the `src` dir shouldn't be included.

pmg/views.py
```diff
@@ -78,25 +78,26 @@ def classify_attachments(files):
...
+            cte_meetings = load_from_api('v2/committees/%s/meetings' % committee.id,
+                fields=['id','title','date','committee_id'])['results']
 
-            recent_meetings.extend(meetings)
+            recent_meetings.extend(cte_meetings)
 
-        sorted(meetings,key=lambda meeting: meeting['date'])
-        user_following.sort(key=lambda x: x.name, reverse=True)
-        default_meetings = load_from_api('v2/committee-meetings/', fields=['id','title','date'])['results'][:10]
+        sorted(recent_meetings,key=lambda meeting: meeting['date'])
```
longhotsummer: This doesn't change recent_meetings, I think you may want `recent_meetings.sort()`

pmg/views.py
```diff
@@ -76,6 +76,31 @@ def classify_attachments(files):
...
+    default_meetings = load_from_api('v2/committee-meetings/', fields=['id','title','date','committee_id'])['results'][:10]
+    default_committees = load_from_api('v2/committees', return_everything=True, fields=['id','name','date'])['results'][:20]
+
+    if current_user.is_authenticated():
+        # Append user-followed committees if logged in
+        user_following = current_user.following
+        recent_meetings = []
+        cte_meetings = []
+
+        for committee in user_following:
```
longhotsummer: Some users follow 100+ committees. This is going to be too slow. We should use the same technique is in `user_followed_committee_meetings()` -- perhaps move that into a helper method on the user object itself, such as `recent_meetings_for_followed_committees`.

pmg/user_management.py
```diff
@@ -62,8 +66,45 @@ def user_committee_alert(committee_id):
         ga_event('user', 'add-alert', 'cte-alert-box')
         flash("We'll send you email alerts for updates on this committee.", 'success')
 
-    return redirect(request.values.get('next', '/'))
+    return redirect(request.headers.get('referer', '/'))
+
+@app.route('/user/follow/committee/<int:committee_id>', methods=['POST'])
+def user_follow_committee(committee_id):
+    if current_user.is_authenticated() and request.method == 'POST':
+        u = current_user.follow_committee(Committee.query.get(committee_id))
```
longhotsummer: we can trim this down quite a bit:

```python
current_user.follow_committee(Committee.query.get(committee_id))
db.session.commit()
```

pmg/models/users.py
```diff
@@ -149,19 +149,25 @@ def gets_alerts_for(self, committee):
             committee = Committee.query.get(committee)
         return committee in self.committee_alerts
 
+    def get_followed_committee_meetings(self):
+        from ..models.resources import CommitteeMeeting
+        from ..api.schemas import CommitteeMeetingSchema
+        following = CommitteeMeeting.committee_id.in_([f.id for f in current_user.following])
+        meetings = CommitteeMeeting.query.filter(following).order_by(desc(CommitteeMeeting.date)).limit(10)
+
+        return CommitteeMeetingSchema(many=True, only=['id', 'title', 'date']).dump(meetings).data
```
longhotsummer: Ah, `.data` :)

You're not always going to want to return this as json, though. Just return the `meetings` object, then it's up to the caller to determine what to do with it. In the one case we can send it as-is to the template, in the other cases it needs to be JSON-ified.

Otherwise we're mixing the business logic of "find the recent meetings that this user cares about" with the view logic of "I need it in this form to show it".

pmg/models/users.py
```diff
@@ -149,19 +149,25 @@ def gets_alerts_for(self, committee):
             committee = Committee.query.get(committee)
         return committee in self.committee_alerts
 
+    def get_followed_committee_meetings(self):
+        from ..models.resources import CommitteeMeeting
+        from ..api.schemas import CommitteeMeetingSchema
+        following = CommitteeMeeting.committee_id.in_([f.id for f in current_user.following])
```
longhotsummer: `current_user` should be `self`

src/pip-delete-this-directory.txt
longhotsummer: This file and everything under src needs to be removed.

pmg/models/users.py
```diff
@@ -228,6 +245,23 @@ def subscribe_to_newsletter(user):
         db.Integer(),
         db.ForeignKey('committee.id', ondelete="CASCADE")))
 
+user_following = db.Table(
```
smmbll: i think greg and i feel that this isn't necessary.

pmg/templates/layout.html
smmbll: resolved.

pmg/models/users.py
smmbll: resolved.

src/pip-delete-this-directory.txt
smmbll: will be deleted in final push.

pmg/models/users.py
```diff
@@ -146,6 +149,20 @@ def gets_alerts_for(self, committee):
             committee = Committee.query.get(committee)
         return committee in self.committee_alerts
 
+    def follow_committee(self, committee):
+        from ..models.resources import Committee
+        if not isinstance(committee, Committee):
+            committee = Committee.query.get(committee)
+        self.following.append(committee)
+        return self
```
smmbll: resolved.

pmg/models/users.py
```diff
@@ -146,6 +149,20 @@ def gets_alerts_for(self, committee):
...
+
+    def unfollow_committee(self, committee):
+        from ..models.resources import Committee
+        if not isinstance(committee, Committee):
+            committee = Committee.query.get(committee)
+        self.following.remove(committee)
+        return self
```
smmbll: resolved.

src/pip-delete-this-directory.txt
longhotsummer: It needs to be part of the pull request, this might help: http://stackoverflow.com/questions/1143796/remove-a-file-from-a-git-repository-without-deleting-it-from-the-local-filesyste

longhotsummer:
Ah, much simpler! However, the menu isn't being inserted into the correct container, it lands up having some of the nav item links in it.

screen shot 2016-11-22 at 5 57 56 pm

PR #169, opened 2016-11-22
Unify megamenu shadow and button hover

longhotsummer:
This makes the menu button and the mega menu look like they belong together.

@guushoekman


longhotsummer:
screen shot 2016-11-22 at 6 20 04 pm

PR #172, opened 2016-12-01
Simplify megamenu

longhotsummer:
* remove unnecessary megamenu styles and code that we weren't using
* nest megamenu css since we're using SCSS and it makes it much easier to understand what's going on
* style elements like `h3` inside the megamenu by just styling `.megamenu h3` rather than having to add specific classes to each `h3` element
* remove support for mobile since we don't use it
* make the html valid (only permitted child of `ul` is `li`)

@guushoekman

### DIFF
--- a/pmg/static/resources/css/style.scss
+++ b/pmg/static/resources/css/style.scss
@@ -1246,22 +1246,6 @@ li.committee.item.list-megamenu {
   bottom: -3px;
 }
 
-.menu-mobile {
-    display: none;
-    padding: 20px;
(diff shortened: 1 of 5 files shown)


longhotsummer:
I landed up doing this because the 'twitter' and 'facebook' buttons weren't working due to how the megamenu is laid out. That still isn't actually fixed, but I needed to trim down the html & css to work out what was going on.

I do have an idea of how to fix the social buttons, I just need to get to it.

PR #176, opened 2017-02-06
move social buttons in top menu so they're not hidden

guushoekman:


pmg/templates/_top_links.html
```diff
@@ -9,4 +9,12 @@
   <li><a href="/bills">Bills</a></li>
   <li><a href="/question_replies">Questions and Replies</a></li>
   <li><a href="/calls-for-comments">Calls for Comments</a></li>
+  <div class="pull-right hidden-xs hidden-sm">
```
longhotsummer: This <div> should be an <li>

guushoekman:
The padding looks random but it's to right-align the buttons with the login / user button.

The CSS is applied to `.btn` in `nav.links`. Is that alright or should these have their own class?

pmg/static/resources/css/style.scss
```diff
@@ -111,6 +111,9 @@ nav.top-links {
 
   .btn {
     margin-top: 6px;
+    padding: 5px 10px;
+    margin-left: 9px;
+    margin-right: -5px;
```
longhotsummer: Doesn't this break any buttons in the mega menu? I'd add a `.social-buttons` class to the LI and style on `.social-buttons .btn`

guushoekman:
Didn't break anything but yeah that makes more sense. Done.

longhotsummer:
Looks good!

PR #182, opened 2017-02-09
Attendance

guushoekman:


pmg/templates/committee_meeting.html
```diff
@@ -77,6 +77,32 @@ <h5 class="date">{{ event.date | pretty_date("long") }}</h5>
       {{ related_bills(event.bills) }}
     {% endif %}
 
+    <div class="panel panel-default attendance">
+      <div class="panel-heading">
+        <h4 class="panel-title">Present</h4>
+      </div>
+      <div>
+        <ul class="list-unstyled list-spaced mp-list">
```
longhotsummer: I don't think the `mp-list` class has any styles linked to it, but I think you're about to add styling for it for the listing on the MP page.

longhotsummer:
Looks good!

PR #189, opened 2017-02-14
Add VAT amounts to subscription fees

xybrnet:
@longhotsummer


longhotsummer:
Looks good, just asking Gaile for input.

PR #191, opened 2017-02-14
Listen for change on cteList and cteListSearchResults

xybrnet:
@longhotsummer

This fixes the issue of not being able to follow a committee when searched for.


longhotsummer:
👍

PR #192, opened 2017-02-14
Deactivate megamenu

xybrnet:
@longhotsummer


longhotsummer:
Looks good. How does it play with the checkboxes for cte following, if at all?

xybrnet:
On the cte listing page? It works fine.
The only thing that has been removed are the event listeners on the menu item.

longhotsummer:
Cool, merge and deploy.

PR #195, opened 2017-02-16
Show all committees a user follows under 'Committees you follow' tab

xybrnet:
@longhotsummer

### DIFF
--- a/pmg/templates/committee_list.html
+++ b/pmg/templates/committee_list.html
@@ -160,26 +160,28 @@
     {% if current_user.is_authenticated() %}
     <div class="tab-pane row cte-list-user-following" id="user-following">
       {% if user_follows_committees %}
-      <ul class="col-xs-12">
-        {% for committee in user_following %}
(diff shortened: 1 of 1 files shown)


pmg/templates/committee_list.html
```diff
@@ -160,26 +160,28 @@ <h3 class="no-results">No results found.</h3>
...
-              <input type="checkbox" checked />
-              <label>
-                <a href="{{ url_for('committee_detail', committee_id=committee.id) }}" class="name">{{ committee.name }}</a>
-                {% if committee.premium %}
-                <span class="premium exclude"><i class="fa fa-key"></i> <a href="{{ url_for('committee_subscriptions') }}"></a></span>
-                {% endif %}
-              </label>
-            </form>
-          </li>
+        {% for column in current_user.following|slice(2) %}
```
longhotsummer: Does that slice() do the right thing? Just looks weird.

xybrnet: http://jinja.pocoo.org/docs/2.9/templates/#slice

PR #201, opened 2017-04-19
Add new API endpoint, meetings-by-member

xybrnet:
@longhotsummer 

API endpoint for attendance per meeting, by member.

### DIFF
--- a/pmg/api/v1.py
+++ b/pmg/api/v1.py
@@ -518,7 +518,7 @@ def committee_meeting_attendance_meetings_by_member():
             m = members.get(member_id, None)
             if m:
                 member = build_attendance_member_dict(m)
-                meetings = [{'attendance': row.attendance, 'date': row.meeting_date} for row in member_rows]
+                meetings = [{'meeting_id': row.meeting_id, 'attendance': row.attendance, 'date': row.meeting_date} for row in member_rows]
 
                 meetings_by_member.append({
                     'member': member,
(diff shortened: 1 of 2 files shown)


longhotsummer:
@xybrnet can you give a quick example of the output of the API, and why the existing APIs don't work? Is this for all MPs, over all time?

xybrnet:
@longhotsummer 

`{
  "member":{
    "party_name":"DA",
    "pa_url":"http://www.pa.org.za/person/david-john-maynier/",
    "party_id":2,
    "id":2,
    "name":"Maynier,` Mr D"},
  "meetings":[
    {"date":"2017-01-31","attendance":"A"},
    {"date":"2017-01-24","attendance":"P"},
    {"date":"2017-01-25","attendance":"P"}]}
`

We need the date for each meeting, so that we can determine the position of the member on the meeting date. The summary endpoint, only give us aggregates for number of meetings present, absent, arrived late etc. for each member.

It does get all the attendance records, for all time. I used the same query that existed for the summary API to get the members. But not all the members end up being served over the API.

longhotsummer:
Makes sense, thanks. Can you include the meeting ID in the attendance detail dict? That gives a useful foreign key. 

Greg
--
(iPhone)

xybrnet:
Cool, I thought that may be useful too.

longhotsummer:
LGTM

PR #206, opened 2017-04-25
adding year dropdown below list

guushoekman:
- second dropdown below
- added jquery so it moves up when changing the bottom dropdown and the two dropdowns match value


longhotsummer:
Nice.

PR #207, opened 2017-04-26
bill fix, pretty date

guushoekman:


pmg/views.py
```diff
@@ -234,10 +235,16 @@ def bill(bill_id):
         'introduced': 1,
     }
     history = bill_history(bill)
+
+    if bill['status']:
```
jbothma: to fix a key error for `bill['status']` do `if 'status' in bill` - this is still a key error and it's only testing whether `bill['status']` is falsy

jbothma: To fix TypeError here (because `bill['status']` exists and the value is `None`, you could do 
```python
if bill.get('status', None):
```
which says get status, but if it's not present in the dict, default to None.

jbothma: so if it doesn't exist in the dict, or it's None or empty string, you'll go to the else clause

PR #208, opened 2017-04-26
V2 api supports premium content

longhotsummer:
* before this, we were leaking premium content in the v2 api
* make the PREMIUM_FREE_BEFORE date a config option


longhotsummer:
@guushoekman when you're rendering the committee meeting page, you can check for `premium_but_free` which will be True if this is premium content but it's now been made freely available.

PR #210, opened 2017-04-26
Social sharing

guushoekman:


pmg/static/resources/javascript/pmg.js
```diff
@@ -145,3 +145,26 @@ $(function() {
     }
   });
 });
+
+$(function() {
+  $(".page-share.facebook").on("click", function(e) {
+    e.preventDefault();
+    window.open("https://www.facebook.com/sharer/sharer.php?u="+window.location.href,
+      "share", "width=600, height=400, scrollbars=no");
+      ga('send', 'social', 'facebook', 'share', url);
```
longhotsummer: The `url` variable isn't set. You probably want something like this:

```javascript
url = window.location.href;
window.open(..., url, ...);
ga('send', 'social', 'facebook', 'share', url);
```

ditto for the lines below, too.

PR #234, opened 2017-07-06
Cte detail filter bugfix

guushoekman:
Fix for https://www.pivotaltracker.com/story/show/145025793


longhotsummer:
What was the underlying cause, here?

guushoekman:
See comments [here](https://www.pivotaltracker.com/story/show/145025793). Those filters were never meant to be there, I added them accidentally.

longhotsummer:
:) ship it

PR #235, opened 2017-07-06
Bug fix for when committee question has no committee

guushoekman:
https://www.pivotaltracker.com/story/show/144286285


longhotsummer:
This is a reasonable workaround for now. Can we force it to hide the "follow this cte" button?

PR #240, opened 2017-08-14
add padding when no membership info for cte

guushoekman:
https://www.pivotaltracker.com/story/show/148653105


guushoekman:
https://www.pivotaltracker.com/story/show/148653105

PR #242, opened 2017-08-15
Require email confirmation

guushoekman:
https://www.pivotaltracker.com/story/show/150217815

***before deploy, run this SQL***
```
update "user" set confirmed_at = '2017-08-15';
```

### DIFF
--- a/pmg/models/users.py
+++ b/pmg/models/users.py
@@ -113,6 +113,10 @@ class User(db.Model, UserMixin):
     def __unicode__(self):
         return unicode(self.email)
 
+    @property
+    def confirmed(self):
+        return self.confirmed_at is not None
+
     def has_expired(self):
         return (self.expiry is not None) and (datetime.date.today() > self.expiry)
 
@@ -194,7 +198,7 @@ class User(db.Model, UserMixin):
         tmp.pop('current_login_ip')
         tmp.pop('last_login_at')
         tmp.pop('current_login_at')
-        tmp['confirmed'] = tmp.pop('confirmed_at') is not None
+        tmp.pop('confirmed_at')
         tmp.pop('login_count')
         tmp['has_expired'] = self.has_expired()
 
--- a/pmg/models/users.py
+++ b/pmg/models/users.py
@@ -117,8 +117,7 @@ class User(db.Model, UserMixin):
     def __unicode__(self):
         return unicode(self.email)
 
-    @property
-    def confirmed(self):
+    def is_confirmed(self):
         return self.confirmed_at is not None
 
     def has_expired(self):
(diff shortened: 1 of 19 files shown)


pmg/models/users.py
```diff
@@ -113,6 +117,10 @@ class User(db.Model, UserMixin):
     def __unicode__(self):
         return unicode(self.email)
 
+    @property
+    def confirmed(self):
```
longhotsummer: Existing methods use naming like `is_authenticated` and `has_expired`, perhaps `is_confirmed` would be a more consistent name? The former two are also methods, rather than properties.

pmg/templates/security/email/welcome.txt
```diff
@@ -2,12 +2,9 @@
 {% block content %}
 
 Welcome to PMG and thanks for signing up.
-{% if security.confirmable %}
-Please confirm your email by clicking here {{ confirmation_link }}
-{% endif %}
 
-To sign up for free email alerts and get the daily parliamentary schedule and alerts when committees are updated, visit https://pmg.org.za/email-alerts/
+Please confirm your email by clicking this link:
```
longhotsummer: "email address".

pmg/templates/user_management/email_alerts.html
```diff
@@ -25,39 +25,60 @@
...
     </div>
   {% endif %}
+  {% if current_user.is_authenticated() and not current_user.confirmed %}
+    <div class="row">
+      <div class="col-sm-5 col-sm-offset-4">
+        <div class="alert alert-info">
+          <h4>Confirm your email address</h4>
+          <p>Once you confirm your email address, you can subscribe to free email alerts.</p>
+          <br>
+          <p>You were sent an email with confirmation link when you signed up. If you have not received it, you can <a href="{{ url_for('security.send_confirmation', next=request.full_path) }}">request another confirmation email</a>.</p>
```
longhotsummer: Perhaps "We sent you an email with a confirmation link..." -- active voice is better than passive.

pmg/templates/user_management/email_alerts.html
```diff
@@ -25,39 +25,60 @@
...
-      </label>
-    </li>
-  </ul>
-
-  <div>
-    <!-- Nav tabs -->
-    <ul class="nav nav-tabs" role="tablist">
-      <li role="presentation" class="active"><a href="#committee-alerts" aria-controls="commitee-alerts" role="tab" data-toggle="tab">Committee Alerts</a></li>
-      <li role="presentation"><a href="#search-alerts" aria-controls="search-alerts" role="tab" data-toggle="tab">Search Alerts</a></li>
+  <fieldset{% if not current_user.confirmed %} disabled{% endif %}>
```
longhotsummer: clever :)

longhotsummer:
Looks good to me, just minor comments about wording.

pmg/models/users.py
jbothma: the to_dict property that this is bringing to the backend was already called `confirmed` and I didn't want to change the API (largely because I'm lazy, but also cos I'm not sure who might be using this). Do you think it's safe to change the API?

PR #258, opened 2018-01-09
fix attendance and smarter years

guushoekman:
- Temporary fix for attendance when there are no meetings yet
- Text on attendance page doesn't need to be updated by year anymore, an update on `views.py` will update that too

### DIFF
--- a/pmg/views.py
+++ b/pmg/views.py
@@ -334,7 +334,9 @@ def attendance_overview():
     """
     Display overview of attendance for meetings.
     """
-    this_year = datetime.today().year
+    # this_year = datetime.today().year
+    this_year = 2017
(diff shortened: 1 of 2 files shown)


pmg/views.py
```diff
@@ -334,7 +334,9 @@ def attendance_overview():
     """
     Display overview of attendance for meetings.
     """
-    this_year = datetime.today().year
+    # this_year = datetime.today().year
```
longhotsummer: Nice. Maybe future proof against ourselves. Keep `this_year` as it is, then:

```python
month = datetime.today().month
# in jan and feb only, show previous year's attendance data
if month < 3:
    this_year = this_year -1
```

PR #272, opened 2018-01-25
2018 attendance in overview

guushoekman:
I've removed the 3 month logic from the attendance overview as I couldn't see how we could include a form of logic while also showing 2018 already. Perhaps we can add the logic for 1 month after January? That way it will remain but we'll also prevent it breaking early next year.

Added a `data-value` for the attendance column. For some reason the sorting wasn't working properly.


longhotsummer:
Looks good.

PR #274, opened 2018-03-08
updated to 15% VAT

guushoekman:


longhotsummer:
I'm guessing we hold off on this until 1 April or end of March?

guushoekman:
Ah is that when the VAT change happens? If so then yes. I'll set a reminder in Slack so we can merge this then.

PR #283, opened 2018-03-15
Committee bills

longhotsummer:
@guushoekman this creates an api `v2/committees/32/bills` which lists all bills linked to a committee.


longhotsummer:
@guushoekman do you need this still?

guushoekman:
Ah, no I don't think so. The bills are working fine in the committee details page. There are some differences between `master` and this branch, but everything seems to be working fine.

PR #289, opened 2018-04-23
Add caching to the v1 generic resource list/show API

jbothma:
@cache.memoize caches a view function based on its arguments. It's important that it's around the thing that retuns the view result. If it's above @api.route it will cache its result instead of the view result which is not what we want.

pmg.no_cache is named after nginx's proxy_no_cache config item. In this case we don't serve from the cache if an authentication header is provided.

We can add similar annotations to other view functions.

### DIFF
--- a/pmg/__init__.py
+++ b/pmg/__init__.py
@@ -5,6 +5,7 @@ import os
 from flask import Flask
 from flask.ext.sqlalchemy import SQLAlchemy
 from flask.ext.migrate import Migrate
+from flask_caching import Cache
 from flask_wtf.csrf import CsrfProtect
 from flask_mail import Mail
 from flask_marshmallow import Marshmallow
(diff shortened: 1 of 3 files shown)


pmg/__init__.py
```diff
@@ -22,6 +23,22 @@
     logging.config.dictConfig(yaml.load(f))
 
 
+# Setup Caching
+
+cache = Cache(app, config={
+    'CACHE_TYPE': 'filesystem',
+    'CACHE_DIR': '/tmp/pmg-cache',
+    'CACHE_DEFAULT_TIMEOUT': 60*60,
```
longhotsummer: this is in secs, so 1 hour, right?

pmg/api/v1.py
```diff
@@ -315,6 +315,7 @@ def committee_list():
 @api.route('/<string:resource>/', )
 @api.route('/<string:resource>/<int:resource_id>/', )
 @load_user()
+@cache.memoize(unless=lambda: no_cache(request))
```
longhotsummer: I'd suggest renaming `no_cache` to `should_skip_cache` or similar. `no_cache` sounds like it should be used to prevent things from being cached.

pmg/__init__.py
jbothma: yeah

pmg/api/v1.py
jbothma: ah right - sure

PR #295, opened 2018-04-24
Hansard by house and year

guushoekman:


pmg/static/resources/javascript/hansards.js
```diff
@@ -0,0 +1,19 @@
+// message if no hansards in filter
+if($(".hansards ul.hansard-list > li").length == 0)
+  $(".hansards ul.hansard-list").append("<li>No hansards found. Please expand your search.</li>");
```
jbothma: this is better done in the template than in javascript. We should try to do page content in the template wherever possible.

pmg/static/resources/javascript/hansards.js
```diff
@@ -0,0 +1,19 @@
...
+// select correct URL when selecting house or year
+$(".select-house select, .select-year form input").change(function() {
+  var filterHouse = "";
+  var filterYear = "";
+  var selectedHouse = $(".select-house select").val();
+  var selectedYear = $(".select-year input:checked").attr("year");
+  var ampersand = ""
+  if(selectedHouse) {
+    var filterHouse = "filter[house_id]=";
+        ampersand = "&";
```
jbothma: I forgot jquery has a way to do this https://stackoverflow.com/a/12216561/1305080

params =  {}
so we'd do something like if(selectedYear != all) { params['filter[year]'] = selectedYear }
and a similar if statement for additional filters. Then only the chosen filters get added to the url, and the filter fields get urlencoded automatically I think.

jbothma: that would also remove the ?all which isn't really needed to achieve what we want.

PR #296, opened 2018-05-03
Show inline PDFs for bills

longhotsummer:
Shows inline PDFs in the bill page. Click-to-load on mobile to save data.

This includes a bit of a hack to allow requesting pdf by CORS via pmg.org.za since S3 doesn't provide CORS headers.
pmg test_5000_bill_208_


longhotsummer:
I don’t see why not. Won’t hurt s3 and enables fun things. Need to apply retroactively. 

Greg
--
(iPhone)

PR #301, opened 2018-06-20
Subscribe to committee flow

xybrnet:
@longhotsummer

### DIFF
--- a/pmg/static/resources/css/style.scss
+++ b/pmg/static/resources/css/style.scss
@@ -1941,18 +1941,3 @@ ul.menu {
     animation: fa-spin 1s infinite steps(16);
   }
 }
-
-.link-btn {
-    background: none;
-    border: none;
-    color :$link-color;
-    text-decoration: none;
-    cursor: pointer;
-    padding: 0;
-    font-size: $font-size-h3;
- 
-    &:hover {
-      text-decoration: underline;
-      color: darken($link-color, 15%);
-    }
-}
--- a/pmg/static/resources/css/style.scss
+++ b/pmg/static/resources/css/style.scss
@@ -1941,3 +1941,18 @@ ul.menu {
     animation: fa-spin 1s infinite steps(16);
   }
 }
+
+.link-btn {
+    background: none;
+    border: none;
+    color :$link-color;
+    text-decoration: none;
+    cursor: pointer;
+    padding: 0;
+    font-size: $font-size-h3;
+ 
+    &:hover {
+      text-decoration: underline;
+      color: darken($link-color, 15%);
+    }
+}
--- a/pmg/templates/security/login_user.html
+++ b/pmg/templates/security/login_user.html
@@ -61,11 +61,14 @@
 
       <div class="row">
         <div class="col-sm-6 col-sm-offset-3">
-          <p class="lead">
+          <h3>
             <a href="{{ url_for('security.forgot_password') }}">I forgot my password.</a>
-            <br>
-            <a href="{{ url_for('security.register') }}{% if login_user_form.next %}?next={{ login_user_form.next.data }}{% endif %}">Don't have an account? Sign up.</a>
-          </p>
+          </h3>
+          <form action="{{ url_for('security.register') }}" method="POST">
+              <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+              <input type="hidden" name="next" value="{{ login_user_form.next.data }}">
+              <input class="link-btn" name="login" type="submit" value="Don't have an account? Sign up.">   
+          </form>
         </div>
       </div>
     </div>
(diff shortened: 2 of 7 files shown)


pmg/user_management.py
```diff
@@ -196,3 +196,19 @@ def get_megamenu():
...
+        current_user.follow_committee(committee)
+
+    if committee not in current_user.committee_alerts:
+        current_user.committee_alerts.append(committee)
+    
+    db.session.commit()
+
+    ga_event('user', 'follow-committee', 'cte-follow-committee')
+
+    return
```
xybrnet: Is this the best place to put this function?
Should it rather return something specific in case of success or failure?

pmg/views.py
```diff
@@ -345,6 +346,18 @@ def committee_detail(committee_id):
     )
 
 
+@app.route('/committee/<int:committee_id>/follow-cte')
+@app.route('/committee/<int:committee_id>/follow-cte/')
+def committee_detail_follow_cte(committee_id):
+    follow_cte_post_registration_or_login(committee_id)
+    
+    # Remove confirmation email flash
+    session['_flashes'] = []
```
xybrnet: We already show another message at the top of the page mentioning the confirmation email. The page looks very busy with all the banners, so I thought we could remove the login extensions confirmation message.

I'm not sure if this is the best way to do it though?

pmg/templates/security/login_user.html
```diff
@@ -61,11 +61,14 @@ <h1>Login now to view premium content</h1>
...
             <a href="{{ url_for('security.forgot_password') }}">I forgot my password.</a>
-            <br>
-            <a href="{{ url_for('security.register') }}{% if login_user_form.next %}?next={{ login_user_form.next.data }}{% endif %}">Don't have an account? Sign up.</a>
-          </p>
+          </h3>
+          <form action="{{ url_for('security.register') }}" method="POST">
+              <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+              <input type="hidden" name="next" value="{{ login_user_form.next.data }}">
+              <input class="link-btn" name="login" type="submit" value="Don't have an account? Sign up.">   
+          </form>
```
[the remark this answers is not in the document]
xybrnet: There are other instances like this that we should update if this is the correct approach.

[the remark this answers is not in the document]
longhotsummer: This doesn't make sense to me -- this should be part of the URL, going to the sign-up page and preserving the next parameter should be able to be done via URL params. But we could be using flask-security weirdly.

pmg/views.py
```diff
@@ -345,6 +346,18 @@ def committee_detail(committee_id):
...
 
 
+@app.route('/committee/<int:committee_id>/follow-cte')
+@app.route('/committee/<int:committee_id>/follow-cte/')
+def committee_detail_follow_cte(committee_id):
+    follow_committee(committee_id)
+    
+    # Remove confirmation email flash
+    session['_flashes'] = []
+    flash('You have successfully followed this committee to receive email alerts.', 'success')
```
longhotsummer: How about "You're now following this committee and we'll send you email alerts when new content is posted."

pmg/views.py
```diff
@@ -345,6 +346,18 @@ def committee_detail(committee_id):
     )
 
 
+@app.route('/committee/<int:committee_id>/follow-cte')
+@app.route('/committee/<int:committee_id>/follow-cte/')
+def committee_detail_follow_cte(committee_id):
+    follow_committee(committee_id)
+    
+    # Remove confirmation email flash
+    session['_flashes'] = []
```
longhotsummer: Don't we want to show this?

pmg/user_management.py
longhotsummer: Seems okay to me. It doesn't need the final `return` statement.

pmg/static/resources/css/style.scss
```diff
@@ -1941,3 +1941,18 @@ ul.menu {
     animation: fa-spin 1s infinite steps(16);
   }
 }
+
+.link-btn {
```
longhotsummer: We should be able to use Bootstrap's btn-link?

pmg/views.py
xybrnet: Yes, that's much better.

pmg/views.py
xybrnet: We're already showing a message about it at the top of the page. I just thought it looks rather busy with all the notifications.

screenshot at jun 21 09-36-21

pmg/static/resources/css/style.scss
xybrnet: Cool, I somehow missed that it exists. I'll use that and just style it to match the password reset text.

screenshot at jun 21 09-45-15

jbothma:
@xybrnet looks good to me! shall we deploy this morning then we can keep an eye during the day?

PR #304, opened 2018-07-05
Quoted search must match exactly

jbothma:


pmg/search.py
```diff
@@ -126,18 +125,43 @@ def mapping(self, data_type):
...
+                            "term_vector": "with_positions_offsets",
+                        },
+                    },
                 },
                 "fulltext": {
                     "type": "string",
                     "analyzer": "english",
                     "index_options": "offsets",
                     "term_vector": "with_positions_offsets",
+                    "store": "yes",
```
jbothma: @longhotsummer this is needed to get highlighting on quoted and unquoted with less boilerplate.

Do you have a sense for how much more space this might use? The server has about 2G left so I should probably free some before deploying anyway and monitor as I deploy.

longhotsummer: I don't know. I'd imagine the current index isn't bigger than a GB or so, but it's a guess.

longhotsummer:
LGTM

pmg/search.py
jbothma: I managed to make a lot of space with a bit of apt and docker cleanup so this is no concern any more.

PR #309, opened 2018-07-09
Use date_of_introduction as primary search key for bills

xybrnet:
The list is sorted twice to sort by number and introduction dates in reverse order, and and then on other keys.

Sort by the secondary keys, then the primary key. See https://docs.python.org/2/howto/sorting.html#sort-stability-and-complex-sorts

### DIFF
--- a/pmg/views.py
+++ b/pmg/views.py
@@ -324,10 +324,10 @@ def committee_detail(committee_id):
 
     bills = load_from_api(
         'v2/committees/%s/bills' % committee_id,
-        fields=['id', 'title', 'status', 'date_of_introduction', 'code', 'number'],
+        fields=['id', 'title', 'status', 'date_of_introduction', 'code'],
     )['results']
 
-    bills.sort(key=lambda b: [b['date_of_introduction'], b.get('number', 0)], reverse=True)
+    bills.sort(key=lambda b: b['date_of_introduction'], reverse=True)
 
     return render_template('committee_detail.html',
                            current_year=now.year,
--- a/pmg/views.py
+++ b/pmg/views.py
@@ -208,14 +208,9 @@ def bills(bill_type, year=None):
     api_url = 'bill' if bill_type == 'all' else 'bill/%s' % bill_type
     bills = load_from_api(api_url, return_everything=True, params=params)['results']
 
-    # Sort bills by date_of_introduction, then other attributes.
-    # date_of_introduction == None defaults to the latest
-    bills.sort(key=lambda b: [-b['year'], b['type']['prefix'], b.get('number', 0), b['title']])
-    bills.sort(
-        key=lambda b: [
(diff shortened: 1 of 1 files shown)


pmg/views.py
```diff
@@ -208,7 +208,9 @@ def bills(bill_type, year=None):
     api_url = 'bill' if bill_type == 'all' else 'bill/%s' % bill_type
     bills = load_from_api(api_url, return_everything=True, params=params)['results']
 
-    bills.sort(key=lambda b: [-b['year'], b['type']['prefix'], b.get('number', 0), b['title']])
+    # Sort bills by number, date_of_introduction, and then other attributes.
+    bills.sort(key=lambda b: [-b['year'], b['type']['prefix'], b['title']])
```
longhotsummer: I don't think this will have the desired effect. The second sort will override the first. I suggest:

`bills.sort(key=lambda b: [-b['year'], b['type']['prefix'], b.get('number', 0), b['date_of_introduction'], b['title']])`

This will put the most recent years first, then sort by type, then number, then date (ascending), then break ties (if any) by sorting on the title.

xybrnet: I'm not sure if this will work as we need either. It will give results like this:

screenshot at jul 12 09-59-31

The way I understood it was that the number and date of introduction should take preference (both in reverse order, to put the latest first, and using both, as bills could have its number or d.o.i returned as None).

Because numbers are sometimes None, `b.get('number', 0)` will return None, and we can't use `-b.get('number', 0)`. We need date of intro in reverse too.

The code I had above seemed to be giving the results we're expecting. Am I misunderstanding the docs I linked to above? In sorting twice to sort some keys in reverse and others in ascending order.

longhotsummer:
Trying to sort this out here: https://docs.google.com/document/d/1wSJWqOla4ggtqUCPztw56QgGky__yNnPqT2t0C93OZM/edit#

xybrnet:
The main issue was that we were sorting by the type prefix, which for post 2016 PMB bills had values of PMB, and not the recoded B.

We can sort by the first letter of the recoded value, though this makes some assumptions on future bill types. Is this okay to do, or should we rather return another value when building the bill dict, that we use specifically for sorting?

pmg/views.py
```diff
@@ -207,8 +207,8 @@ def bills(bill_type, year=None):
 
     api_url = 'bill' if bill_type == 'all' else 'bill/%s' % bill_type
     bills = load_from_api(api_url, return_everything=True, params=params)['results']
-
-    bills.sort(key=lambda b: [-b['year'], b['type']['prefix'], b.get('number', 0), b['title']])
+ 
+    bills.sort(key=lambda b: [-b['year'], b['code'][0],  b.get('number', 0), b['title']])
```
longhotsummer: Genius.

PR #317, opened 2018-09-17
Add sphere as attendance query string parameter

xybrnet:
@longhotsummer
 
Filter members by sphere in the query string.
Default to national if not specified.

### DIFF
--- a/pmg/api/v1.py
+++ b/pmg/api/v1.py
@@ -571,7 +571,6 @@ def committee_meeting_attendance_meetings_by_member():
 
 
 @api.route('/committee-meeting-attendance/data.xlsx')
-@api.route('/committee-meeting-attendance/data.xlsx/')
 def committee_meeting_attendance_download():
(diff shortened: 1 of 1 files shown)


pmg/api/v1.py
```diff
@@ -555,6 +571,7 @@ def committee_meeting_attendance_meetings_by_member():
 
 
 @api.route('/committee-meeting-attendance/data.xlsx')
+@api.route('/committee-meeting-attendance/data.xlsx/')
```
longhotsummer: Sure you want the ending slash? That looks weird to me.

longhotsummer:
LGTM. Just a comment on the ending slash.

pmg/api/v1.py
xybrnet: I had an issue earlier where it seemed to drop the query string, but it works.
So, yes, definitely dropping it.

PR #335, opened 2019-05-20
add legacy reports link

ebsuku:


pmg/templates/_footer.html
```diff
@@ -32,6 +32,7 @@
           <li><a href="/committees/">Committees</a></li>
           <li><a href="{{ url_for('daily_schedules') }}">Daily Schedules</a></li>
           <li><a href="/hansards">Hansards</a></li>
+	  <li><a href="/page/LegacyReports">Legacy Reports</a></li>
```
PiDelport: Should we not be using `url_for` here?

```suggestion
	  <li><a href="{{ url_for('page', pagename = 'LegacyReports') }}">Legacy Reports</a></li>
```

ebsuku: ah, yes.
Thanks.

PR #342, opened 2019-06-20
6th parliament committee

ebsuku:
* Non active committees must be greyed out and under the active committees
* Remove 5th Parliament banner


pmg/templates/committee_list.html
```diff
@@ -3,232 +3,289 @@
...
+				    {% endif %}
+				    <label>
+					<a href="{{ url_for('committee_detail', committee_id=committee.id) }}" class="name">{{ committee.name }}</a>
+					{% if committee.premium %}
+					    <span class="premium exclude"><i class="fa fa-key"></i></span>
+					{% endif %}
+				    </label>
+				</form>
+			    </li>
+			{% else %}
```
jbothma: I don't understand why we're duplicating a lot of this HTML now?

pmg/templates/committee_list.html
```diff
@@ -3,232 +3,289 @@
...
+					    <span class="premium exclude"><i class="fa fa-key"></i></span>
+					{% endif %}
+				    </label>
+				</form>
+			    </li>
+			{% endif %}
+		    {% endif %}
+		{% endfor %}
+	    </ul>
+	{% endfor %}
```
jbothma: Are we now not closing the divs created in the for loop?

pmg/templates/committee_list.html
```diff
@@ -3,232 +3,289 @@
...
-    </div>
+	<div class="tab-pane prov cte-items" id="prov">
+	    <div class="row prov-row">
+		{% for key, ctype in wc_committees.iteritems() %}
+		    <div class="col-xs-12 col-md-6 single-prov">
+			<h3>{{ ctype.name }} Parliament</h3>
+			<ul>
+			    {% for committee in ctype.committees %}
+				{% if committee.active %}
+				    <li class="committee item {% if not committee.active %}in{% endif %}active" data-id="{{ committee.id }}" {% if committee.followed %} data-following="true" {% endif %}>
```
jbothma: is this the same markup as above? Can we factor it out to a template tag? That'd make it so much easier to see what's actually happening here

pmg/templates/committee_list.html
```diff
@@ -3,232 +3,289 @@
...
+		{% endfor %}
+	    </div>
+	    <hr>
+	    <div class="row prov-row">
+		{% for key, ctype in wc_committees.iteritems() %}
+		    <div class="col-xs-12 col-md-6 single-prov">
+			<ul>
+			    {% for committee in ctype.committees %}
+				{% if not committee.active %}
+				    <li>
```
jbothma: is this also a repeat of the above? not sure if an include or template tag is more appropriate but some reuse would make the length and indentation/tag closing easier

pmg/templates/committee_list.html
```diff
@@ -3,232 +3,289 @@
...
+					<label>
+					    <a href="{{ url_for('committee_detail', committee_id=committee.id) }}" class="name">
+						<span style="color:grey">{{ committee.name }}</span>
```
jbothma: wouldn't a bootstrap disabled class or custom css in our css file be better than inline style?

jbothma: like text-muted used above

ebsuku:
I just took another look at this.
One of the reasons why I choose to duplicate some stuff, was for readability.
I've extracted some parts that could be abstracted.

PR #344, opened 2019-06-25
Attendance overview

ebsuku:
We are simply moving the old parliaments attendance data to a new page.
Once new attendance data comes in can we then correctly setup the attendance page for the new parliament.

### DIFF
--- a/pmg/models/resources.py
+++ b/pmg/models/resources.py
@@ -1297,10 +1297,7 @@ class CommitteeMeetingAttendance(ApiResource, db.Model):
         return rows
 
     @classmethod
-    def annual_attendance_trends(cls,
-                                 from_year=None,
-                                 to_year=None,
-                                 period=None):
+    def annual_attendance_trends(cls, to_year=None, period=None):
         """ Attendance summary by year and committee. Excludes ad-hoc committees.
         Returns row tuples: (committe_id, house name, year, n_meetings, avg_attendance, avg_members)
         """
--- a/pmg/models/resources.py
+++ b/pmg/models/resources.py
@@ -1360,7 +1360,7 @@ class CommitteeMeetingAttendance(ApiResource, db.Model):
 
     @classmethod
     def annual_attendance_rank_for_committee(cls, committee, year):
-        attendance = cls.annual_attendance_trends(from_year=year, to_year=year)
+        attendance = cls.annual_attendance_trends(to_year=year)
 
         # match house
         attendance = [a for a in attendance if a.house == committee.house.name_short]
--- a/pmg/models/resources.py
+++ b/pmg/models/resources.py
@@ -1297,12 +1297,19 @@ class CommitteeMeetingAttendance(ApiResource, db.Model):
         return rows
 
     @classmethod
-    def annual_attendance_trends(cls, from_year, to_year):
+    def annual_attendance_trends(cls,
+                                 from_year=None,
+                                 to_year=None,
+                                 period=None):
         """ Attendance summary by year and committee. Excludes ad-hoc committees.
         Returns row tuples: (committe_id, house name, year, n_meetings, avg_attendance, avg_members)
         """
-        start_date = datetime.datetime(from_year, 1, 1)
-        end_date = datetime.datetime(to_year, 12, 31)
+        if period == 'historical':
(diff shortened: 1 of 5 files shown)


pmg/templates/attendance_overview.html
```diff
@@ -4,101 +4,111 @@
...
-        </div>
-        <div class="value-mobile hidden-md hidden-lg">
-          <span class="label label-{% if item.change > 0 %}success{% else %}danger{% endif %}">{% if item.change > 0 %}+{% endif %}{{ item.change|int }}%</span>
-        </div>
-      {% endif %}
-      </td>
-    </tr>
-    {% endfor %}
-  </table>
+                              <thead>
```
jbothma: did anything actualy change here?
Also why's it now suuuuuper indented?

pmg/views.py
```diff
@@ -407,19 +407,18 @@ def committee_detail_follow_cte(committee_id):
...
-def attendance_overview():
+def archived_attendance_overview():
     """
-    Display overview of attendance for meetings.
+    Show archived attendance view
     """
-    month = datetime.today().month
-    this_year = datetime.today().year
-    last_year = this_year - 1
+    this_year = 2019
```
jbothma: how are we ensuring we fix this? Is there a story that is very clearly something that can't be forgotte int he backlog?

jbothma: oh wait, this is archived - so for the next 5 years these years should work jsut fine?

pmg/models/resources.py
```diff
@@ -1297,12 +1297,19 @@ def meetings_by_member(cls):
         return rows
 
     @classmethod
-    def annual_attendance_trends(cls, from_year, to_year):
+    def annual_attendance_trends(cls,
+                                 from_year=None,
```
jbothma: from_year and to_year aren't used any more, start_date and end_date are hardcoded - this will be very confusing

pmg/views.py
ebsuku: Correct, by archived only shows for the previous year, which doesn't really make sense anymore. It should be an archive for the entire history of the parliament. So we'll need to write something for that, I've added a Pivotal for that.

jbothma:
Confirmed, this is not trying to show new data automatically. It's just hiding old data.

We need to make further changes to support showing new data.

pmg/templates/attendance_overview.html
jbothma: Now this is spaces but still super indented?

pmg/templates/attendance_overview.html
```diff
@@ -4,101 +4,111 @@
...
+  {% if not attendance %}
+  <h4>Data for current parliament is not avaliable yet.</h4>
+  {% else %}
   <table class="table table-hover sortable">
-    <thead>
-      <th class="th-rank"></th>
-      <th class="th-cte-name">Committee</th>
-      <th class="th-number-meetings hidden-xs" data-firstsort="desc">Meetings</th>
-      <th class="th-attendance" data-defaultsort="desc">Attendance</th>
-      <th class="th-attendance-change">Since {{ year - 1}}</th>
```
jbothma: This will be wrong in 6 months - we need to make this dynamic now or ensure we fix it statically or dynamically so it doesn't show the wrong thing in January

pmg/templates/attendance_overview.html
```diff
@@ -4,101 +4,111 @@
...
-      </td>
-    </tr>
-    {% endfor %}
-  </table>
+                              <thead>
+	                          <th class="th-rank"></th>
+	                          <th class="th-cte-name">Committee</th>
+	                          <th class="th-number-meetings hidden-xs" data-firstsort="desc">Meetings</th>
+	                          <th class="th-attendance" data-defaultsort="desc">Attendance</th>
+	                          <th class="th-attendance-change">Since {{ year }}</th>
```
jbothma:
```suggestion
	                          <th class="th-attendance-change">Since 2019</th>
```

pmg/templates/attendance_overview.html
jbothma: updated this to reflect what's needed https://www.pivotaltracker.com/story/show/166897271

PR #346, opened 2019-06-30
Ministries

ebsuku:
* Add a notification showing that a committee/ministry is inactive.
* If a committee is not active hide the follow this committee box on committee page

### DIFF
--- a/pmg/templates/_active_committee_detail.html
+++ b/pmg/templates/_active_committee_detail.html
@@ -1,5 +1,5 @@
-{% if not committee.active %}
-<div class="alert alert-warning">
-    <strong>Warning!</strong> This ministry is now inactive. Find new questions under the ministry housing this function now
+{% if not committee.active and committee.minister %}
+<div class="alert alert-info">
+    <strong>Notice:</strong> This ministry is now inactive. Find new questions under the ministry housing this function now.
 </div>
 {% endif %}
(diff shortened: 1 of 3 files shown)


pmg/templates/_active_committee_detail.html
```diff
@@ -0,0 +1,5 @@
+{% if not committee.active %}
+<div class="alert alert-warning">
```
jbothma: info, not warning

pmg/templates/_active_committee_detail.html
```diff
@@ -0,0 +1,5 @@
...
+    <strong>Warning!</strong> This ministry is now inactive. Find new questions under the ministry housing this function now
```
jbothma:
```suggestion
    <strong>Notice:</strong> This ministry is now inactive. Find new questions under the ministry housing this function now
```

pmg/templates/committee_detail.html
```diff
@@ -117,7 +118,8 @@ <h3 class="no-results">No results.</h3>
           {% endif %}
         </div>
         <div class="tab-pane fade in cte-dtl-meetings-tab-rep" id="tr">
-          {% if committee.tabled_committee_reports %}
+	   {% include '_active_committee_detail.html' %}
```
jbothma: the notice above is specific to calls and the ministry while the rest of the tabs seem to be non-question content about committees instead - let's do more appropriate wording for them

pmg/templates/_active_committee_detail.html
```diff
@@ -0,0 +1,5 @@
+{% if not committee.active %}
```
jbothma: need to ensure this only shows for committees related to a ministry

PR #347, opened 2019-06-30
Bug attendance data

ebsuku:
Correct the downloading of data on the attendance-overview page

### DIFF
--- a/tests/unit/test_attendance.py
+++ b/tests/unit/test_attendance.py
@@ -96,3 +96,15 @@ class TestCommitteeMeetingAttendance(PMGTestCase):
 
         self.assertEqual([(2019.0, 1L, 0.5, 2.0)], current_attendance)
         self.assertEqual([(2019.0, 2L, 0.5, 2.0)], historical_attendance)
+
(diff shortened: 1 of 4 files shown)


tests/unit/test_attendance.py
```diff
@@ -96,3 +96,15 @@ def test_committee_attendance_trends(self):
...
         self.assertEqual([(2019.0, 2L, 0.5, 2.0)], historical_attendance)
+
+    def test_committee_attendance_summary(self):
+        """
+        checking that the correct number of rows are returned from a attendance summary
+        """
+        current_attendance = CommitteeMeetingAttendance.summary()
+        historical_attendance = CommitteeMeetingAttendance.summary("historical")
+        self.assertEqual([(1, 'P', 2019.0, 1, 1L), (2, 'A', 2019.0, 1, 1L)], current_attendance)
+        self.assertEqual([(1, 'A', 2019.0, 1, 1L),
```
jbothma: Is this order guaranteed? If not you might need to do `assertIn` for each row

jbothma: or something like that

PR #349, opened 2019-07-01
correct deletion of committee

ebsuku:


pmg/templates/admin/my_edit.html
```diff
@@ -8,7 +8,7 @@
         <a href="{{ admin_view.alert_url(model) }}" class="btn btn-default"><i class="fa fa-envelope"></i> Send alert</a>
         {% endif %}
 	{% if admin_view.name == "Committees" %}
-	<script defer src="https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
```
jbothma: Why are we deleting defer here?

ebsuku: The jquery was not loading in time, good ol' $ is not defined.

Found a better way to do it.

pmg/templates/admin/my_edit.html
```diff
@@ -31,10 +30,10 @@
 		</div>
 	    </div>
 	</div>
-	<script defer>
-	 $(document).ready(function(){
+	<script>
+	 window.onload = function(){
```
jbothma: Use `window.addEventListener("load",...` instead, otherwise this might replace another event handler and we'll scratch out heads for ever.

https://stackoverflow.com/a/16683250/1305080

ebsuku: Can you further explain. Its not clear on SO

PR #354, opened 2019-07-04
WIP: Member layout for NA and NCOP

ebsuku:


ebsuku:
member_before_1

member_before

pmg/views.py
```diff
@@ -961,9 +961,13 @@ def members():
                 'sphere'] == 'national':
             members_by_house.setdefault(member['house']['name'],
                                         []).append(member)
+    id_mapping = {
+        house: house.lower().replace(" ", "-") for house in members_by_house.keys()
```
jbothma: isn't there a nice slugify function at hand so you don't need to slugify manually?

ebsuku: Nah, flask doesn't have one.
There are python libraries that can do it but that would mean installing a dependency.

jbothma: ja, I've used https://github.com/un33k/python-slugify - I think adding this as a dep is better than a brittle hack - we don't want this to break because someone decided to put something random into a name

PR #355, opened 2019-07-08
Fix committees in private bill

ebsuku:
Committee bills now have their own endpoint.

Bills are now combined in the view rather than in the bill/pmb query

### DIFF
--- a/pmg/models/resources.py
+++ b/pmg/models/resources.py
@@ -192,10 +192,6 @@ class Bill(ApiResource, db.Model):
     def list(cls):
         return cls.query.order_by(desc(cls.year))
 
-    @classmethod
-    def committee_bill(cls):
-        return cls.query.filter(cls.introduced_by.like("%Committee%"))
-
 
 class BillVersion(db.Model):
     __tablename__ = "bill_versions"
--- a/pmg/models/resources.py
+++ b/pmg/models/resources.py
@@ -192,6 +192,10 @@ class Bill(ApiResource, db.Model):
     def list(cls):
         return cls.query.order_by(desc(cls.year))
 
+    @classmethod
+    def committee_bill(cls):
+        return cls.query.filter(cls.introduced_by.like("%Committee%"))
+
 
 class BillVersion(db.Model):
     __tablename__ = "bill_versions"
--- a/pmg/views.py
+++ b/pmg/views.py
@@ -209,9 +209,9 @@ def bills_explained():
 
 
 @app.route("/bills/<any(current):bill_type>/")
-@app.route("/bills/<any(all, draft, pmb, tabled, committee):bill_type>/")
+@app.route("/bills/<any(all, draft, pmb, tabled, 'pmb-committee'):bill_type>/")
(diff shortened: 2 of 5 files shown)


pmg/models/resources.py
```diff
@@ -192,6 +192,10 @@ def __unicode__(self):
     def list(cls):
         return cls.query.order_by(desc(cls.year))
 
+    @classmethod
+    def committee_bill(cls):
```
jbothma: I think this should be plural - it returns all the matching bills, right?

ebsuku: They get combined in the view template , not in the API

ebsuku: Committee has its own api endpoint.

pmg/views.py
```diff
@@ -230,11 +232,17 @@ def bills(bill_type, year=None):
...
+
+    if bill_type == "pmb":
+        api_url = "bill/committee"
+        committee = load_from_api(api_url, return_everything=True, params=params)[
+            "results"
+        ]
+        bills = bills + committee
 
-    bills.sort(key=lambda b: [-b['year'], b['code'][0],  b.get('number', 0), b['title']])
+    bills.sort(key=lambda b: [-b["year"], b["code"][0], b.get("number", 0), b["title"]])
```
jbothma: - it's not really bill_type, more like bill_view
- keeping the web URL as /bills/pmb but showing pmb and committee bills is confusing but not the end of the world
- formatting in `pmb` for the first lookup and tagging `committee` bills on as a special case just compounds all the issues into something that goes over my threshold what's readable code.

I think fixing the slug is sufficient, because it forces the bill lookup for the "pmb and committees" case to be more explicit.

Here's how I would do it

```python
@app.route('/bills/<any(current):bills_view>/')
@app.route('/bills/<any(all, draft, pmb, tabled):bills_view>/') # redirect to year version - next line
@app.route('/bills/<any(all, draft, pmb, tabled):bills_view>/year/<int:year>/')
def bills(bills_view, year=None):
    if bills_view == 'current':
        # don't paginate by year
        year_list = None
        params = {}

    else:
        year_list = range(MIN_YEAR, date.today().year + 1)
        year_list.reverse()
        params = {}

        if not year:
            return redirect(
                url_for('bills', bills_view=bills_view, year=year_list[0]))

        if year not in year_list:
            abort(404)
        params['filter[year]'] = year

    if bills_view == 'all':
        api_url = 'bill'
        bills = load_from_api(
            api_url, return_everything=True, params=params)['results']
    elif bills_view == 'pmb-commitee':
        api_url = 'bill/pmb'
        pmb = load_from_api(
            api_url, return_everything=True, params=params)['results']
        api_url = 'bill/committee'
        committee = load_from_api(
            api_url, return_everything=True, params=params)['results']
        # merge
        bills = pmb + committee
    else:  # current, draft, tabled
        api_url = 'bill/%s' % bills_view
        bills = load_from_api(
            api_url, return_everything=True, params=params)['results']
```

PR #358, opened 2019-07-18
Linking hansards to various types of bill event titles

ebsuku:


ebsuku:
This is my thinking behind the linking.

Essentially if there is a matching hansard, we append the Id of the hansard to the bill event with the appropriate title.

This way we don't have to modify the entire structure of the bill object.
So a bill event will be link if it has a hansard and will just be static text if there is none.

Downside is that, we have to traverse the entire bill object to check for this.

pmg/bills.py
```diff
@@ -111,9 +111,61 @@ def bill_history(bill):
...
+    Match bill title against the following possible titles
+    """
+    bill_titles = [
+        "Bill passed by National Assembly",
+        "Bill passed by both Houses",
+        "Bill revived on this date",
+        "The NCOP rescinded",
+        "Bill remitted",
+    ]
+    for title in bill_titles:
```
jbothma: another way of writing this is with `any(lambda ...)`
https://docs.python.org/3/library/functions.html#any

Not a change request, just a hint for when it makes sense. It's clear enough here

pmg/bills.py
```diff
@@ -111,9 +111,61 @@ def bill_history(bill):
...
+    for class_history in bill_history:
+        for event_history in class_history["events"]:
+            if event_history["type"] == "house":
+                for event in event_history["events"]:
+                    if event["type"] == "plenary":
+                        for bill_event in event_history["events"]:
+                            if match_title(bill_event["title"]) and match_dates(
+                                event["date"], bill_event["date"]
+                            ):
+                                bill_event["hansard"] = {"id": event["id"]}
```
jbothma: why make a new dict instead of just assigning the event itself?

pmg/bills.py
```diff
@@ -111,9 +111,61 @@ def bill_history(bill):
...
+    The Date of the bill event and the hansard will be the same.
+    Bill Titles we are looking for:
+    * Bill passed by National Assembly
+    * Bill passed by both Houses
+    * Bill revived on this date
+    * The NCOP rescinded
+    * Bill remitted
+    """
+
+    for class_history in bill_history:
```
jbothma: I would say it is important to unit test this - 7 levels of nesting is not something that's quick to verify by eye, but writing the tests should be really easy.

pmg/bills.py
```diff
@@ -111,9 +111,61 @@ def bill_history(bill):
...
+
+def match_dates(hansard_date, event_date):
+    hansard_iso_date = iso8601.parse_date(hansard_date)
+    event_iso_date = iso8601.parse_date(event_date)
+    if hansard_iso_date.date() == event_iso_date.date():
+        return True
+    return False
+
+
+def hansard_linking(bill_history):
```
jbothma: a better name might be "link_hansard" - the imperative rather than the concept

pmg/bills.py
```diff
@@ -111,9 +111,61 @@ def bill_history(bill):
...
+    """
+    We need to link certain bill events to hansards
```
jbothma: I think it's worth documenting:

- does this modify one or more events? Does it modify al events that meet the criteria?
- Does this do anything special if there are more than one appropriate hansards or does it just pick _some_ matching hansard without warning?

pmg/templates/bills/detail.html
```diff
@@ -93,14 +93,8 @@ <h5 class="media-heading">
                         {% if event.type != 'plenary' %}
                         <div class="col-xs-4 col-md-2 text-muted">{{ event.date|pretty_date("long") }}</div>
                         <div class="col-xs-8 col-md-10">
-                            {% if 'Bill passed by National Assembly'  in event.title or 'Bill passed by both Houses' in event.title %}
```
jbothma: love code deletion :heart:

pmg/bills.py
jbothma: I think an integration test where this doesn't have any effect and another where it does would also be in order, so that the thread where this works on objects coming from the API for reals isn't accidentally broken. Doesn't need to check detail logic, just enough to be sure that the integration isn't broken.

PR #359, opened 2019-07-19
Bug fix attendance overview

ebsuku:
Attendance Overview page now showing new parliament stats.

attendance_before

attendance_after

### DIFF
--- a/pmg/templates/attendance_overview.html
+++ b/pmg/templates/attendance_overview.html
@@ -17,26 +17,18 @@
   <h4>Data for current parliament is not avaliable yet.</h4>
   {% else %}
   <table class="table table-hover sortable">
-                              <thead>
-                                  <th class="th-rank"></th>
-                                  <th class="th-cte-name">Committee</th>
-                                  <th class="th-number-meetings hidden-xs" data-firstsort="desc">Meetings</th>
-                                  <th class="th-attendance" data-defaultsort="desc">Attendance</th>
+      <thead>
+          <th class="th-rank"></th>
+          <th class="th-cte-name">Committee</th>
+          <th class="th-number-meetings hidden-xs" data-firstsort="desc">Meetings</th>
+          <th class="th-attendance" data-defaultsort="desc">Attendance</th>
+          {% if total_years > 1 %}
+          <th class="th-attendance-change">Since {{year}}</th>
+          {% endif %}
       </thead>
-                              {% for item in attendance_na %}
-                              <tr id="{{ item.committee_id }}">
-                                  <td class="rank">#{{ loop.index }}</td>
-                                  <td class="cte-name"><a href="{{ url_for('committee_detail', committee_id=item.committee_id) }}">{{ item.committee }}</a></td>
-                                  <td class="number-meetings hidden-xs">{{ item.n_meetings }}</td>
-                                  <td class="attendance" data-value="{{ item.avg_attendance }}">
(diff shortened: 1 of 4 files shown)


pmg/templates/attendance_overview.html
```diff
@@ -13,98 +13,56 @@ <h1><i class="fa fa-fw fa-table" aria-hidden="true"></i> Committee meeting atten
...
                                       <div class="progress">
-		                          <div class="progress-bar progress-bar-success" role="progressbar" aria-valuenow="{{ item.avg_attendance }}" aria-valuemin="0" aria-valuemax="100" style="width: {{ item.avg_attendance }}%;">
-		                              {{ item.avg_attendance|round|int }}%
-		  </div>
+                                          <div class="progress-bar progress-bar-success" role="progressbar" aria-valuenow="{{ item.avg_attendance }}" aria-valuemin="0" aria-valuemax="100" style="width: {{ item.avg_attendance }}%;">
+                                              {{ item.avg_attendance|round|int }}%
+                  </div>
               </div>
-	  </td>
-	                          <td class="attendance-change" data-value="{{ item.change }}">
```
jbothma: I don't think the solution is to remove a feature that has to be re-added before the end of the year.
I think the solution is to add a test so we know it works, and then add a condition to hide the column when the information isn't available yet.

If that's hard because the backend is currently hardcoding cut-off dates, you can add a template-only test with mocked context data.

pmg/views.py
```diff
@@ -491,10 +491,50 @@ def attendance_overview():
     Display overview of attendance for meetings.
     """
     this_year = datetime.today().year
-    attendance = CommitteeMeetingAttendance.annual_attendance_trends(
-        to_year=this_year)
+    attendance = CommitteeMeetingAttendance.annual_attendance_trends(to_year=this_year)
```
jbothma: We had a bug which led to something not working when we were confident it would and told the client it's done.

Manual inspection wasn't sufficient here - let's add the automated testing needed to be confident we won't need to talk to the client about this being broken again.

pmg/templates/attendance_overview.html
ebsuku: After looking at this again. I think this is the way to go.
We will be making a change(hiding the 2019 column) that effects this page once every 5 years for a few months.
To get that column back would involve ideally a git revert.
Thoughts?

pmg/views.py
ebsuku: The issue was that we had not implemented the attendance for the current data since it was a bit tricky dealing with the upcoming 6 months.

pmg/templates/attendance_overview.html
jbothma: The reason I'm suggesting we add a test and dynamically hide this is that ongoing maintenance is amost guaranteed to mean that it won't be as simple as a git revert.

By keeping the feature in the page and having a test, we're forced to keep it in mind with ongoing maintenance.

Most importantly, it's working right now. Including a test of this should take like an hour.
Fixing it when it turns out it's not just a git revert will take more than an hour. and either means we have to remember to check for it coming up to next year, or it will mean stress because it turns out not to work.

If testing for it working and making it work is so hard, it suggests there's a problem with our ability to test for stuff, which needs to be fixed anyway.

pmg/views.py
jbothma: There's all this code that isn't tested?

PR #366, opened 2019-08-12
Fix attendance overview

delenamalan:


tests/unit/test_attendance.py
```diff
@@ -47,7 +47,7 @@ def setUp(self):
             attendance='P',
             member=jabu,
             meeting=old_parliament_meeting,
-            created_at='2019-01-01')
+            created_at='2019-08-01')
```
jbothma: why did you change this? Was it to make the test fail to confirm the bug before fixing it?

delenamalan: Both. It's a valid edge case, is it not?

PR #374, opened 2019-08-21
Improve Blog page: easier to find previous posts and ability to add image to the side bar

delenamalan:


pmg/static/resources/javascript/bills.js
```diff
@@ -68,5 +68,6 @@ $(function() {
 
   // show the first tab when the page loads
   var firstTab = $('.bill-version-content a[data-toggle="tab"]')[0];
-  autoShowBillPdf(firstTab.getAttribute('href').slice(1));
+  if (firstTab)
```
delenamalan: Reason for this change is the console error that I got that prevented any Javascript defined after `bills.js` to be run.

Screenshot at 2019-08-21 20-45-23

pmg/templates/blog_post.html
```diff
@@ -8,8 +8,13 @@
   <div class="row">
     <div class="col-sm-8">
       <div class="single-post">
+        <div class="date">{{ post.date.strftime('%d %B %Y') }}</div>
         <h2 class="title">{{ post.title }}</h2>
-        <div class="date">{{ post.date.strftime('%d-%m-%Y') }}</div>
+        <div class="preview-image">
+            {% if post.get_preview_image() %}
+              <img src="{{ post.get_preview_image().url }}" alt="{{ post.get_preview_image().title }}" />
```
delenamalan: It feels wrong calling `get_preview_image` three times, but I'm not sure what else to do here?

I could return the image url (or the placeholder url if the image does not exist) from `get_preview_image`, but then we'll lose the image's `title` property.

pmg/views.py
```diff
@@ -1701,20 +1701,59 @@ def correct_this_page():
...
+
+    years = db.session.query(
+                cast(months.c.year, Integer).label('year'), 
+                func.sum(months.c.month_posts).label('year_posts'), 
+                func.array_agg(months.c.month).label('months'), 
+                func.array_agg(months.c.month_name).label('month_names'), 
+                func.array_agg(months.c.month_posts).label('months_posts'))\
+        .group_by('year')\
+        .order_by(year.desc())\
+        .all()
```
delenamalan: Returns something like:
```
[{
 'month_names': [u'May'], 
 'months': [5.0], 
 'months_posts': [1L],
 'year_posts': Decimal('1'),
 'year': 2023
}, {
 'month_names': [u'May', u'September', u'August', u'January'], 
 'months': [5.0, 9.0, 8.0, 1.0],
 'months_posts': [1L, 1L, 12L, 1L],
 'year_posts': Decimal('15'),
 'year': 2019
}]
```

Where each index in `month_names`, `months` and `months_posts` corresponds to each other. I.e. `month_names[0]` is the month's name for `month[0]` and `months_posts[0]` is the number of posts in `month[0]`.

I could figure out how to rather return a single array for those three properties combined instead, e.g. `[{ 'month_name': 'May', 'month': 5, 'months_posts': 1}, ...]`.

pmg/templates/blog_post.html
jbothma: there might be "assign" semantics in jinja but let's see how this scales and improve on it if really necessary. It's not like it's going to hit the DB three times (I'd assume)

delenamalan: Ah looks like there is indeed an "assign" semantic. Will try it out!

tests/views/test_blog_pages.py
```diff
@@ -0,0 +1,87 @@
...
+        super(TestBlogPages, self).tearDown()
+
+    def test_blog_post_page(self):
+        """
+        Test blog post page (http://pmg.test:5000/blog/<slug>)
+        """
+        post = self.fx.PostData.the_week_ahead
+        self.get_page_contents("http://pmg.test:5000/blog/%s/" % post.slug)
+        self.assertIn(post.title, self.html)
+        self.assertIn(post.body[0:100], self.html)
```
jbothma: I tend to recommend hardcoding strings in the asserts rather than using what's in the fixture data programmatically since it'll be easier to notice the test becoming invalid if someone changes the value in the fixture to something like the empty string if it's hardcoded. It's very little duplication with a big win in robustness.

jbothma: I don't think you need to go and change all this now. If it's quick, that'd be nice but this is more for the future. I'll try and capture this in best practises as well.

delenamalan: Ah ok. So like:

`self.assertIn("The Week Ahead: End of the First Term", self.html)`?

tests/views/test_blog_pages.py
```diff
@@ -0,0 +1,87 @@
...
+        self.assertIn('That week in Parliament', self.html)
+        self.assertIn('About this blog', self.html)
+        self.assertIn('Blog Archive', self.html)
+
+    def contains_posts(self, posts):
+        for post in posts:
+            self.contains_blog_post(post)
+
+    def contains_blog_post(self, post):
+        self.assertIn(flask.escape(post.title), self.html)
```
jbothma: The comment on hardcoding doesn't make so much sense here, but perhaps the arguments to this and contains_posts could be a list of strings and IDs or something

PR #376, opened 2019-08-23
Run correct elastic index script on travis and dev

delenamalan:
The `index_elastic.py` script that was added when we dockerised PMG doesn't index our models correctly. It doesn't use our transforms and mappings defined in `pmg/search.py` which breaks search on a develop (Docker) environment. We can rather use the script that was used before to index the models.


pmg/search.py
```diff
@@ -406,7 +406,10 @@ def reindex_everything(self):
             self.reindex_all(data_type)
 
     def delete_everything(self):
-        self.es.delete_index(self.index_name)
+        try:
+            self.es.delete_index(self.index_name)
+        except ElasticHttpNotFoundError:
+            pass
```
jbothma: catching and dropping an exception without logging it can make debugging incredibly difficult. It's almost never the right thing to do.

In what circumstances could this occur? If it's reasonable to expect it to occur but it doesn't acually affect anything, perhaps an INFO level log is appropriate.

If it actually indicates an error, perhaps an exception log including stack trace is appropriate.

delenamalan: The exception happens the first time `python bin/search.py --reindex all` command is run for on a new host, because the `pmg` index won't exist yet. I'll add an INFO-level log message.

PR #377, opened 2019-08-26
"bill passing" event matching for linking to "hansard events" not matching all needed

delenamalan:
https://www.pivotaltracker.com/n/projects/1367366/stories/168057198

This is what the error message on the admin page would look like:
Screenshot at 2019-08-26 17-16-43


pmg/__init__.py
```diff
@@ -193,7 +193,7 @@ def send_email_with_sendgrid(message):
 
 # if we don't do this in a separate thread, we hang trying to connect to the db
 import pmg.tasks
-Thread(target=pmg.tasks.schedule).start()
+# Thread(target=pmg.tasks.schedule).start()
```
jbothma: is this supposed to be commented out still?

delenamalan: Sorry no, committed that by mistake.

delenamalan:
I see the "help" drop down doesn't work for new events. Will look into that now.

PR #380, opened 2019-08-30
questions and replies should be paginated by years

delenamalan:
https://www.pivotaltracker.com/n/projects/1367366/stories/155809025


jbothma:
Why does sports arts and culture show when I've selected transport?

Screenshot_2019-09-02_11-58-10

PR #381, opened 2019-09-16
Move attendance 6th term epoch to start of June (instead of July)

delenamalan:


delenamalan:
@jbothma hold on before merging, please. I think there's some more changes that need to be made.

delenamalan:
Fixed now. Ready to merge.

PR #389, opened 2020-01-15
Update dependencies

delenamalan:
https://www.pivotaltracker.com/n/projects/1367366


jbothma:
Please fix these page types and ensure there's a smoke test for them (like just a simple thing loading the page and roughly checking a string unique to that page is present in the response (and fails when this error is shown

Screenshot_2020-01-16_07-50-41
Screenshot_2020-01-16_07-45-22
Screenshot_2020-01-16_07-45-42

jbothma:
Please also click through to the other attendance-reated pages so we can be sure they work properly - perhaps it's just the data download view and the archived attendance page

jbothma:
Ignore my comments - all good when I rebuild with new deps

PR #396, opened 2020-02-09
Fix CKEditor bugs

delenamalan:
This PR fixes the errors reported about the HTML editor elements in the admin site.

Somehow CKEditor instances were created before our [admin.js](https://github.com/OpenUpSA/pmg-cms-2/blob/master/pmg/static/resources/javascript/admin/admin.js#L3) script was supposed to create them. There must be a selector somewhere that creates a CKEditor instance for all elements with the `ckeditor` class, but I can't find it anywhere (checked in FlaskAdmin as well).

So I changed the class name of all of the admin fields we want to be CKEditors to `ckeditor_instance` so that it won't be created before we want it to be created with our settings.

Our settings include the `autogrow` plugin which when we don't have it installed, causes the second error reported and `disallowedContent: 'a[!name]'` which absence causes anchors to be created from bookmarks pasted from Word documents. 

[Pivotal](https://www.pivotaltracker.com/n/projects/1367366/stories/171251893)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -86,4 +86,4 @@ XlsxWriter==0.6.6
 xlrd==1.0.0
 # Heroku (thus dokku) recomments editable if from git and relying on on upstream changes
 # https://devcenter.heroku.com/articles/python-pip#git-backed-distributions
--e git+https://github.com/Code4SA/za-parliament-scrapers.git@master#egg=za_parliament_scrapers
+-e git+https://github.com/Code4SA/za-parliament-scrapers.git@c9a4a1f4f02152bd2a7c25977a843e2d5e2fe656#egg=za_parliament_scrapers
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -86,4 +86,4 @@ XlsxWriter==0.6.6
 xlrd==1.0.0
 # Heroku (thus dokku) recomments editable if from git and relying on on upstream changes
 # https://devcenter.heroku.com/articles/python-pip#git-backed-distributions
--e git+https://github.com/Code4SA/za-parliament-scrapers.git@master#egg=za_parliament_scrapers
+-e git+https://github.com/Code4SA/za-parliament-scrapers.git@c9a4a1f4f02152bd2a7c25977a843e2d5e2fe656#egg=za_parliament_scrapers
```
jbothma: good idea!

jbothma:
Nice catch - please change the class to `pmg-ckeditor` though, since the `instance` suffix doesn't really tell me "load our custom settings ckeditor". Then you can ship

PR #398, opened 2020-02-20
Upgrade to python 3 again

delenamalan:
Fix the errors that we experienced when we upgraded earlier today, namely:

- Error on search pages
- Error on files pages, e.g. https://pmg.org.za/files/file.pdf
- Error on https://pmg.org.za/question_replies
- Error on https://pmg.org.za/committee/111/


pmg/utils.py
```diff
@@ -48,14 +48,15 @@ def track_pageview(path=None, ignore_bots=True):
...
 
-    tracker = Tracker.create(ga_id, user_id=user_id, client_id=client_id)
-    tracker.send(
-        "pageview",
-        path,
-        uip=request.access_route[0],
-        referrer=request.referrer or "",
-        userAgent=user_agent,
-    )
+    with HTTPRequest() as http:
```
jbothma: shouldn't this be the request that triggered the call rather than a blank new request?

jbothma: I know it'll take a bit more wiring to pass the request along, but if I'm not mistaken, this will give the impression that every logged request is a new user, totally messing up the data.

How did it work before? Did it get the request from the context?

delenamalan: TBH I didn't think about what this is actually for. The Python 2 [version](https://github.com/adswerve/universal-analytics-python) of this package didn't require the `http` parameter. I'll look into to find out what how it's used.

jbothma:
Is there a test covering the search , question/reply and committee issues or are we adding them later?

delenamalan:
- [This commit](https://github.com/OpenUpSA/pmg-cms-2/pull/398/commits/e806f1a145258d8f782a88aa937f7c31726da29d) recreates the error that happened on the search results pages.
- [This commit](https://github.com/OpenUpSA/pmg-cms-2/pull/398/commits/44f6a725428c2aa69ed4d10600a91e682f385824) adds a fixture that recreates the question-replies error.
-  The fixture added in [this commit](https://github.com/OpenUpSA/pmg-cms-2/pull/398/commits/cee2615e44c6ed0f1a20bd1746824dcc0abaf292) covers the committee page error.

pmg/utils.py
delenamalan: Looks like the old package created a new request using the user agent if there was one:

https://github.com/adswerve/universal-analytics-python/blob/78ee36a5f1efc144043ae003c855b97fbe662f7c/UniversalAnalytics/Tracker.py#L217-L218

jbothma: Ah cool - ok, this isn't the time to be improving these things - so we're just maintaining behaviour. I'll take a peak at the analytics in a day or so just to be sure.

jbothma:
Love it. Ship it!

PR #399, opened 2020-02-20
Fix python 3 errors

delenamalan:
Fix errors on https://pmg.org.za/calls-for-comments/?filter%5Bcommittee%5D=37 and https://api.pmg.org.za/committee-meeting-attendance/data.xlsx

### DIFF
--- a/pmg/models/soundcloud_track.py
+++ b/pmg/models/soundcloud_track.py
@@ -11,10 +11,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 SOUNDCLOUD_ARTWORK_PATH = "pmg/static/resources/images/logo-artwork.png"
-SOUNDCLOUD_AUTH_500_MSG = (
-    "500 Server Error: Internal Server Error for url:"
-    " https://api.soundcloud.com/oauth2/token"
-)
+SOUNDCLOUD_AUTH_500_MSG = "500 Server Error: Internal Server Error for url"
 UNFINISHED_STATES = [
     "storing",
     "stored",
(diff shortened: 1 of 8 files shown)


pmg/models/soundcloud_track.py
```diff
@@ -11,10 +11,7 @@
 logger = logging.getLogger(__name__)
 
 SOUNDCLOUD_ARTWORK_PATH = "pmg/static/resources/images/logo-artwork.png"
-SOUNDCLOUD_AUTH_500_MSG = (
-    "500 Server Error: Internal Server Error for url:"
-    " https://api.soundcloud.com/oauth2/token"
-)
+SOUNDCLOUD_AUTH_500_MSG = "500 Server Error: Internal Server Error for url"
```
jbothma: This is relaxing the constraint - I think it's fine to ignore 500s for that URL, but not for all. So let's keep the URL in there.

tests/api/test_committee_meeting_attendance_excel_export.py
```diff
@@ -0,0 +1,21 @@
...
+    def tearDown(self):
+        self.fx.teardown()
+        super(TestCommitteeMeetingAttendanceExcelExport, self).tearDown()
+
+    def test_get_committee_meeting_attendance_xlsx_smoke_test(self):
+        res = self.client.get(
+            "committee-meeting-attendance/data.xlsx",
+            base_url="http://api.pmg.test:5000/",
+        )
+        self.assertEqual(200, res.status_code)
```
jbothma: perhaps an easy improvement to this test is to just open it with openpyx or whatever the library is already in this project to see if it fails. The simplest operation that fails if the response is _not_ an excel file.

jbothma:
Looks good to me!

PR #400, opened 2020-02-21
Bill page event sorting error

jbothma:
https://stackoverflow.com/questions/55695479/typeerror-not-supported-between-instances-of-dict-and-dict


jbothma:
looks good to me!

Please mail Rashaad when verified fixed

PR #401, opened 2020-02-24
Fix confirm email

delenamalan:
[Sentry error](https://sentry.io/organizations/openupsa/issues/1524041671/?project=1489874&query=is%3Aunresolved)

### DIFF
--- a/tests/unit/test_sharpspring.py
+++ b/tests/unit/test_sharpspring.py
@@ -17,10 +17,28 @@ class MockResponse:
         return self.json_data
 
 
-def mocked_requests_post_success(*args, **kwargs):
+def get_requests_post_success(*args, **kwargs):
     return MockResponse(
         {"result": {"creates": [{"success": True,}]}, "error": False}, 200
     )
+
+
+def get_sharpspring_call_success(*args, **kwargs):
+    return {"result": {"creates": [{"success": True,}]}, "error": False}
+
+
+def get_sharpspring_call_fail_create_lead(*args, **kwargs):
+    return {"result": {"creates": [{"success": False, "error": {"code": 500}}]}}
+
+
+def get_sharpspring_call_fail_add_list_member(method, params):
+    if method == "createLeads":
+        return {"result": {"creates": [{"success": True,}]}, "error": False}
+    else:
+        return {
+            "result": {"creates": [{"success": False, "error": {"code": 500}}]},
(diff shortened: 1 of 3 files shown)


tests/unit/test_sharpspring.py
```diff
@@ -0,0 +1,102 @@
...
+        {"result": {"creates": [{"success": True,}]}, "error": False}, 200
+    )
+
+
+def get_sharpspring_call_success(*args, **kwargs):
+    return {"result": {"creates": [{"success": True,}]}, "error": False}
+
+
+def get_sharpspring_call_fail_create_lead(*args, **kwargs):
+    return {"result": {"creates": [{"success": False, "error": {"code": 500}}]}}
```
jbothma: does this not raise in raise_for_status?

tests/unit/test_sharpspring.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def get_sharpspring_call_fail_add_list_member(method, params):
+    if method == "createLeads":
+        return {"result": {"creates": [{"success": True,}]}, "error": False}
+    else:
+        return {
+            "result": {"creates": [{"success": False, "error": {"code": 500}}]},
```
jbothma: ditto

tests/unit/test_sharpspring.py
delenamalan: My mock response doesn't: https://github.com/OpenUpSA/pmg-cms-2/pull/401/files#diff-90beb5472c712ae3ffaa178e243ca4c9R13-R14

Since the sharpspring code checks for error codes I'm assuming there must be instances where `raise_for_status` doesn't raise an error, but there is still some error code. It probably won't be a 500 code, but I haven't looked at the Sharpspring docs or anything to see what it might be. 

https://github.com/OpenUpSA/pmg-cms-2/pull/401/files#diff-ca4d79d8ac81dcd09cb9a1137ab1fcfcR47

delenamalan: Looks like "Object-level" errors won't raise an error according to the [Sharpspring docs](https://help.sharpspring.com/hc/en-us/articles/115001069228-Understanding-SharpSpring-Open-API-Overview#h_8587766891311518636374378).

jbothma:
See the comments above but regardless, this looks fine - ship it.

PR #405, opened 2020-03-25
Organisation update error

delenamalan:
[Pivotal](https://www.pivotaltracker.com/n/projects/1367366/stories/171647494)

This PR aims to prevent the following error that prevented PMG from updating an organisation:

error

The situation that caused this error is as follows:
1. The organisation was "subscribed" to a committee that isn't a premium committee.
2. A subscription is represented by a record in the `organisation_committee` table in the database.
3. There were duplicate records in the `organisation_committee` for the above subscription, e.g.
```
pmg=# select * from organisation_committee;
 organisation_id | committee_id 
-----------------+--------------
               1 |            1
               1 |            1
               1 |            1
```
4. When you submit the form, Flask-Admin tries to delete all of the subscriptions that aren't ticked in the form.
5. Even though PMG didn't update the subscriptions, Flask-Admin tries delete all of the subscriptions that aren't ticked under "premium subscriptions".
6. Non-premium subscriptions are being hided before displaying the form.
7. So Flask-Admin tries to delete the non-premium subscriptions.
8. SQLAlchemy expects its going to delete one record (per subscription), but because there are duplicate subscriptions, it raises an exception. [This](https://stackoverflow.com/questions/31768652/deleting-from-many-to-many-sql-alchemy-and-postgresql) is a similar issue on StackOverflow.

This PR introduces a unique constraint on the `organisation_id` and `committee_id` columns in the `organisation_committee` table so that duplicate subscriptions won't be allowed in the future. I also included a script and test for the script to delete existing duplicate subscriptions.

I still haven't been able to identify how duplicate subscriptions were created in the first place.

Steps to deploy will be to:
- Merge
- `dokku run python bin/remove_duplicate_subscriptions.py --commit`
- `dokku run python app.py db upgrade`


tests/views/test_admin_organisations_page.py
```diff
@@ -69,3 +70,30 @@ def test_post_admin_organisations_new_page(self):
...
+        url = "/admin/organisation/edit/?id=%d"
+        organisation = self.fx.OrganisationData.pmg
+        self.create_organisation_data["subscriptions"] = self.fx.CommitteeData.arts.id
+        response = self.make_request(
+            url % organisation.id,
+            self.user,
+            data=self.create_organisation_data,
+            method="POST",
+            follow_redirects=True,
+        )
```
jbothma: are we missing an assertion here or is a request without an exception sufficient?

PR #408, opened 2020-04-22
Cascade file deletions to event_files

delenamalan:
When someone deletes a file, the file needs to be automatically unlinked from a event (such as a committee meeting) as well.

[Pivotal](https://www.pivotaltracker.com/story/show/172451735)

### DIFF
--- a/pmg/models/resources.py
+++ b/pmg/models/resources.py
@@ -231,7 +231,7 @@ class File(db.Model):
     origname = db.Column(db.String(255))
     description = db.Column(db.String(255))
     playtime = db.Column(db.String(10))
-    event_files = db.relationship("EventFile", lazy=True)
+    event_files = db.relationship("EventFile", lazy=True, cascade="all, delete")
 
     def to_dict(self, include_related=False):
         tmp = serializers.model_to_dict(self, include_related=include_related)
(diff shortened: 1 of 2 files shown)


pmg/models/resources.py
```diff
@@ -231,7 +231,7 @@ class File(db.Model):
     origname = db.Column(db.String(255))
     description = db.Column(db.String(255))
     playtime = db.Column(db.String(10))
-    event_files = db.relationship("EventFile", lazy=True)
+    event_files = db.relationship("EventFile", lazy=True, cascade="all, delete")
```
delenamalan, line 234: No database migration needed because the database constraint already exists. It is SQLAlchemy that tries to set `file_id` to `NULL` before deleting anything and letting the database handle the cascade.

tests/unit/test_files.py
```diff
@@ -0,0 +1,61 @@
...
+        db.session.add(self.committee)
+        db.session.add(self.committee_meeting)
+        db.session.add(self.event_file)
+        db.session.commit()
+        self.file_id = self.file.id
+
+    @patch("pmg.models.resources.s3_bucket", return_value=MockS3())
+    def test_delete_file_when_linked_to_meeting(self, mock_s3_bucket_key):
+        # When we delete the file, the event should be deleted too, but
+        # the meeting shouldn't be deleted
```
jbothma, line 53:
```suggestion
        # When we delete the file, the event_file join table entry should be deleted too,
        # but the meeting (event) shouldn't be deleted
```

PR #409, opened 2020-05-04
Fix timezone display

delenamalan:
Set the timezone option for Postgres to 'Africa/Johannesburg' so that all timezone-aware times are returned in that zone from Postgres.

From the Postgres [docs](https://www.postgresql.org/docs/8.3/datatype-datetime.html#DATATYPE-TIMEZONES):

> All timezone-aware dates and times are stored internally in UTC. They are converted to local time in the zone specified by the timezone configuration parameter before being displayed to the client.

We also need to use the `LocalDateTime` field in the Marshmallow schemas because the default [`DateTime`](https://marshmallow.readthedocs.io/en/2.x-line/api_reference.html#marshmallow.fields.DateTime) field in Marshmallow < 3.0.0 converts all dates to UTC.

TODO:
- [ ] Use the `LocalDateTime` field in other schemas that contain dates as well.
- [ ] Considering upgrading to Marshmallow > 3.0.0 instead because in newer versions dates [aren't](https://marshmallow.readthedocs.io/en/stable/upgrading.html#datetime-leaves-timezone-information-untouched-during-serialization) automatically converted to UTC.

### DIFF
--- a/docker-compose-test.yml
+++ b/docker-compose-test.yml
@@ -3,4 +3,4 @@ version: '3'
 services:
   web:
     environment:
-      - SQLALCHEMY_DATABASE_URI=postgresql://pmg:pmg@postgres/pmg_test?client_encoding=utf8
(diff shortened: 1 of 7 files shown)


delenamalan:
Closing because we rather want to make the change in the display code.

PR #410, opened 2020-05-04
Fix timezone display (no changes to database connection)

delenamalan:
Fixes the same issue as #409, but doesn't alter the database connection options.

Instead, changes the display logic of Flask Admin and Jinja so ensure that dates always get displayed in South African time.

~~TODO:~~
~~- [ ] Check how the change to `pretty_date` affects other models or move it to a separate filter.~~
~~- [ ] Use the same `on_form_prefill` and `column_formatters` for other models with dates.~~
Will make updates for other models in a next PR.

### DIFF
--- a/pmg/helpers.py
+++ b/pmg/helpers.py
@@ -10,9 +10,12 @@ from flask import request, url_for
 from pmg import app
 from pmg.ga import get_ga_events
 from pmg.forms import CorrectThisPageForm
+import pytz
 
 
 logger = logging.getLogger(__name__)
+
+SAST = pytz.FixedOffset(120)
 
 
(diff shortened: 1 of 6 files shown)


pmg/helpers.py
```diff
@@ -10,10 +10,13 @@
 from pmg import app
 from pmg.ga import get_ga_events
 from pmg.forms import CorrectThisPageForm
+import pytz
 
 
 logger = logging.getLogger(__name__)
 
+SAST = pytz.FixedOffset(120)
```
jbothma, line 18: isn't there something like `pytz.timezone('Africa/Johannesburg')` ?

PR #413, opened 2020-07-20
Fix question upload

delenamalan:
Turns out upgrading BeautifulSoup also fixes the error.

[Pivotal](https://www.pivotaltracker.com/n/projects/1367366/stories/173500049)

Made a [PR](https://github.com/Code4SA/za-parliament-scrapers/pull/4) on ZAParliamentScrapers to set the minimum BeautifulSoup version.


tests/views/test_admin_committee_questions.py
```diff
@@ -124,6 +124,57 @@ def test_upload_committee_question_document_with_new_format(self):
         # Delete the question that was created
         self.created_objects.append(question)
 
+    def test_upload_committee_question_document_with_navigable_string_error(self):
+        """
+        Upload committee question document (/admin/committee-question/upload)
+        """
+        url = "/admin/committee-question/upload"
+        data = {}
+        path = self.get_absolute_file_path("../data/RNW1153-200619.docx")
```
jbothma, line 133: can we put it in a folder with the suite or test case name?

delenamalan: Will do

PR #415, opened 2020-10-08
Run task scheduler in dedicated command

delenamalan:
Move code that starts the task scheduler to a new Flask-Script command and run in separate process so that we don't run it in the same amount of processes that we run the web server in.

[Pivotal](https://www.pivotaltracker.com/n/projects/1367366/stories/174667560)

### DIFF
--- a/pmg/tasks.py
+++ b/pmg/tasks.py
@@ -1,3 +1,6 @@
+from flask import current_app
+from flask_script import Command
+from pmg import db
 import logging
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


pmg/tasks.py
```diff
@@ -56,8 +56,46 @@ def schedule():
...
+            log.info(
+                "Not starting task scheduler because RUN_PERIODIC_TASKS is not 'true'"
+            )
+            return
+
+        from apscheduler.schedulers.blocking import BlockingScheduler
+        from apscheduler.jobstores.sqlalchemy import SQLAlchemyJobStore
+        import pmg.tasks
+
+        scheduler = BlockingScheduler(
```
delenamalan, line 84: [Docs](https://apscheduler.readthedocs.io/en/stable/modules/schedulers/blocking.html?highlight=BlockingScheduler#apscheduler.schedulers.blocking.BlockingScheduler) for `BlockingScheduler`.

jbothma (approved):
Nice! Thanks

PR #419, opened 2020-10-26
Add try-catch for Soundcloud sync state

delenamalan:
We're getting a [404](https://sentry.io/organizations/openupsa/issues/1981551299/?project=1489874&query=is%3Aunresolved) for some of the old tracks. Log error for those tracks, but continue processing the next ones.


jbothma (approved):
Just make sure the log message somehow indicates which track it relates to. ID is fine.

PR #423, opened 2020-12-09
Update price list text

delenamalan:
[Pivotal](https://www.pivotaltracker.com/n/projects/1367366/stories/176067699)

The "Premium content from before 2019 is now available for everyone!" text will be automatically updated by the [`PREMIUM_FREE_BEFORE`](https://github.com/OpenUpSA/pmg-cms-2/blob/master/config/config.py#L34-L38) value.


jbothma (approved):
thanks!

PR #443, opened 2021-06-30
Deactivate only

Lunga001:
When meetings are edited for an inactive committee, the committee shouldn't be activated


Lunga001:
Don't activate ad-hoc committee
Committees-Committees-PMG-CMS

When editing committee meeting

Committees-Committee-Meetings-PMG-CMS

PR #496, opened 2023-07-26
Added cache decorator to get_featured_content

desafinadude:


paulmwatson (changes requested):
Because the function has no params it will cache it once and then as long as the cache storage remains it will use the cached result. That could be days or weeks depending. So I think we need to use `cache.delete_memoized(get_featured_content)` somewhere. Either when the relevant database model is changed or a timer that does it every X hours (24?).

paulmwatson (approved):

PR #507, opened 2024-01-17
Remove auth header for internal API requests

michaelglenister:
Story: https://trello.com/c/PqPFnmra/356-pmg-logged-in-user-sql-query-performance


paulmwatson (approved):
Hard to know how much this improves things without running in production. I still see a lot of the `SELECT anon ` SQL queries for every page load. Lets get it deployed and measure improvement.

